PR #11, opened 2017-11-28
Add initial snippets & service page model, expose API

benweatherman:
Implements parts of cityofaustin/joplin#2, cityofaustin/joplin#3, and cityofaustin/joplin#6

Proof is in the pudding:

```shell
curl http://localhost:8000/api/v2/pages/\?format\=json\&type\=base.ServicePage\&fields\=content,extra_content,tags | jq
```

```json
{
  "meta": {
    "total_count": 1
  },
  "items": [
    {
      "id": 4,
      "meta": {
        "type": "base.ServicePage",
        "detail_url": "http://localhost/api/v2/pages/4/",
        "html_url": "http://localhost/sweet-new-page/",
        "slug": "sweet-new-page",
        "first_published_at": "2017-11-28T19:47:05.575059Z"
      },
      "title": "This is a sweet new page",
      "content": "<p>This is some sweet content</p><p><br/></p><ol><li>do this first</li><li>then do this</li><li>now you're done<br/></li></ol>",
      "extra_content": [
        {
          "type": "content",
          "value": "<p><b>This is my extra content</b><br/></p>",
          "id": "695e363a-6ecf-457e-8e53-e351aecc2ad6"
        },
        {
          "type": "application_block",
          "value": 1,
          "id": "dbb9e30f-2aff-4669-aee3-e555192a8580"
        },
        {
          "type": "content",
          "value": "<h1>This is some other extra content<br/></h1>",
          "id": "fb4ab425-9391-496a-aa5f-70915b02bde5"
        }
      ],
      "tags": [
        "nolo",
        "rolo",
        "yolo"
      ]
    }
  ]
}
```


joplin/base/migrations/0003_auto_20171128_1944.py
```diff
@@ -0,0 +1,130 @@
...
+                'abstract': False,
+            },
+        ),
+        migrations.CreateModel(
+            name='Director',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('name', models.CharField(max_length=255)),
+                ('photo', models.CharField(max_length=255)),
+                ('email', models.EmailField(max_length=254)),
```
mateoclarke: don't think this matters at all but why is this magic number different from the rest?

joplin/base/models.py
```diff
@@ -1,25 +1,84 @@
...
-
-from wagtail.api import APIField
+from taggit.models import TaggedItemBase
 
 
 class HomePage(Page):
     pass
 
 
+WYSIWYG_FEATURES = ['h1', 'h2', 'h3', 'bold', 'italic', 'link', 'ul', 'ol']
```
mateoclarke: 👍  Are there any other options in this list you considered or that we may want to include in the future? 

Also wondering if there will be any sanitization or parsing required when passing a string of html to the frontend via the API. There must be a trick to get it to render as HTML. Maybe we'll have to use the `dangerouslySetInnerHtml` method in React: https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml

joplin/base/models.py
```diff
@@ -1,25 +1,84 @@
...
+
+
+class ServicePage(Page):
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    content = RichTextField(features=WYSIWYG_FEATURES, help_text='Write out the steps a resident needs to take to use the service')
```
mateoclarke: 🆒

joplin/base/models.py
```diff
@@ -1,25 +1,84 @@
...
+    panels = [
+        FieldPanel('url'),
+        FieldPanel('description'),
+    ]
+
+    def __str__(self):
+        return self.description
+
+
 @register_snippet
```
mateoclarke: what does this do? what is this syntax?

joplin/base/migrations/0003_auto_20171128_1944.py
benweatherman: You have to manually specify`CharField.max_length` and 255 is a pretty common default. Django chooses a sensible default for `EmailField.max_length` which is apparently 254 :|

joplin/base/models.py
benweatherman: The API responses come back as HTML currently but it's fairly simple to modify if there's a better representation to use.

There are other options available [listed in the docs](http://docs.wagtail.io/en/v1.12/advanced_topics/customisation/page_editing_interface.html#limiting-features-in-a-rich-text-field) but these seemed like most of what an author would need.

joplin/base/models.py
benweatherman: It's a python decorator, which is used to modify a class/function somehow. In this case, wagtail uses it as a way to mark these classes as snippets so they can be used in the admin. Here's more on what wagtail snippets actually are http://docs.wagtail.io/en/latest/topics/snippets.html

mateoclarke: 👍

briaguya:
@benweatherman first thing I noticed:

When hitting the API we're getting HTML out:
```
"content": "<p>This is some sweet content</p><p><br/></p><ol><li>do this first</li><li>then do this</li><li>now you're done<br/></li></ol>",
```
I know we can manage this on the frontend either by parsing out the content from HTML (I'm sure there are some good npm packages for that kind of thing) or by using dangerouslySetInnerHTML in react. I don't think it's a huge issue, but it's definitely something worth thinking about as we build out our components.

For example, instead of having list content coming to us as:
```
<li>do this first</li><li>then do this</li>
```
it'd be nice to get something similar to:
```      
"extra_content": [
        {
          "type": "content",
          "value": "<p><b>This is my extra content</b><br/></p>",
          "id": "695e363a-6ecf-457e-8e53-e351aecc2ad6"
        },
        {
          "type": "application_block",
          "value": 1,
          "id": "dbb9e30f-2aff-4669-aee3-e555192a8580"
        },
        {
          "type": "content",
          "value": "<h1>This is some other extra content<br/></h1>",
          "id": "fb4ab425-9391-496a-aa5f-70915b02bde5"
        }
      ],
```
but with a type of list_item or something of that ilk

I haven't done too much investigation into the wagtail API, but I'm eager to hear any ideas.

joplin/base/migrations/0003_auto_20171128_1944.py
briaguya: I think a decent interim solution here would be to just set pseudo constants in the class:
```
class Migration(migrations.Migration):

    DEFAULT_CHAR_FIELD_MAX = 255
    DEFAULT_EMAIL_FIELD_MAX = 254
```

joplin/base/migrations/0003_auto_20171128_1944.py
```diff
@@ -0,0 +1,130 @@
...
+                'abstract': False,
+            },
+        ),
+        migrations.CreateModel(
+            name='ServicePage',
+            fields=[
+                ('page_ptr', models.OneToOneField(auto_created=True, on_delete=django.db.models.deletion.CASCADE, parent_link=True, primary_key=True, serialize=False, to='wagtailcore.Page')),
+                ('created_at', models.DateTimeField(auto_now_add=True)),
+                ('updated_at', models.DateTimeField(auto_now=True)),
+                ('content', wagtail.wagtailcore.fields.RichTextField(help_text='Write out the steps a resident needs to take to use the service')),
```
briaguya: It'd be cool to implement this as a [streamfield](https://wagtail.io/features/streamfield/) to avoid having the list as just a chunk of HTML

joplin/base/models.py
```diff
@@ -1,25 +1,84 @@
...
 class Event(ClusterableModel):
+    title = models.CharField(max_length=255)
```
briaguya: Maybe we could use multiple inheritance to include pseudoconstants instead of magic numbers here

joplin/base/migrations/0003_auto_20171128_1944.py
benweatherman: These migrations are auto-generated from models, so I'll probably punt on doing that here. I'll add some constants to `models.py` so magic numbers aren't scattered everywhere.

joplin/base/models.py
benweatherman: good suggestion, I'll throw these in as constants

joplin/base/migrations/0003_auto_20171128_1944.py
benweatherman: StreamField has a pretty awkward UI, so I think it would bog down authors with the way they'd have to create new list entries or new paragraphs. I like the idea of separating content into lists. Filed an issue to do better than just returning HTML cityofaustin/joplin#15

benweatherman:
@briaguya I like the suggestion for better return structure. For now, I'll probably keep it so that it generates HTML but I filed an issue to do better cityofaustin/joplin#15 and we can look at that once we're past the initial prototype.

PR #12, opened 2017-11-28
Add fixtures

benweatherman:
Also added a custom logo cause it's a killer feature


benweatherman:
Excellent, thanks for the link to a better logo. I'll use that instead.

I just dumped my local database for now so that there would be _something_ in an API response. I'll modify this soon to include the actual initial data for the service pages but wanted to get the structure in there.

PR #13, opened 2017-11-30
cors config

ifsimicoded:
Basic CORS config for api access on local janis
We'll want to definitely tweak this as we build deeper.


benweatherman:
Thanks for the changes @ifsimicoded!

mateoclarke:
giphy 22

PR #16, opened 2017-12-01
Use fixture data based on real content

benweatherman:
This uses the actual content we're going to start with


benweatherman:
I'm merging because it's just fixture data 😸

PR #29, opened 2017-12-18
Add Locations & Contacts to ServicePage

benweatherman:
The output structure is a little funkë, but it's what we're given by default

```json

{
  "locations": [
    {
      "id": 1,
      "meta": {
        "type": "base.ServicePageLocation"
      },
      "location": {
        "id": 1,
        "meta": {
          "type": "base.Location"
        },
        "name": "Austin Recycle and Reuse Drop-off Center",
        "street": "2514 Business Center Dr",
        "city": "Austin",
        "state": "TX",
        "country": "United States",
        "zip": "78744",
        "hours": [
          {
            "id": 1,
            "meta": {
              "type": "base.LocationDayAndDuration"
            },
            "day_of_week": "Monday",
            "start_time": "08:00:00",
            "end_time": "16:00:00"
          }
        ]
      }
    }
  ],
  "contacts": [
    {
      "id": 1,
      "meta": {
        "type": "base.ServicePageContact"
      },
      "contact": {
        "id": 1,
        "meta": {
          "type": "base.Contact"
        },
        "name": "Francine Hammerschmidt",
        "email": "fran@gmail.com",
        "phone": "5553941212"
      }
    }
  ]
}
```

Also
- Change theme to topic
- Change docker so we can load data from within the image (this a temp change to get fixture data deployed for user testing)


joplin/base/wagtail_hooks.py
```diff
@@ -19,7 +19,11 @@ def editor_css():
 
 @hooks.register('insert_editor_js')
 def editor_js():
-    return format_html('<script src="{}">', static('js/admin.js'))
+    urls = [
+        'https://cdnjs.cloudflare.com/ajax/libs/lodash.js/4.17.4/lodash.min.js',
```
mateoclarke: what are we using lodash for in the admin js? Should we plan to eventually pull in a copy locally?

benweatherman: I haven't made the PR yet, but I'm gonna use it for live preview. This addition accidentally got caught up in this commit.

I'm fairly indifferent in this case about where lodash is served from. I feel like we'll eventually have a legit build step for JS in wagtail because things will get complex enough to warrant it. Thoughts? I'm happy to add a local copy.

PR #33, opened 2017-12-19
Add deploy script for heroku

benweatherman:
Little deets:
- Add necessary `ALLOWED_HOSTS` so django doesn't cry when a request comes from heroku
- Modify Dockerfile to support `PORT` environment variable, which is required by heroku


benweatherman:
I think Brent is checking out the sweet new deployment at https://joplin-staging.herokuapp.com/admin/

PR #55, opened 2018-01-09
Add Department page and endpoint

benweatherman:
You can edit the department via the CMS

edit department

You can query the data via graphql

graphql


benweatherman:
I'm gonna merge to keep the party rolling, but let me know if anyone has feedback

ifsimicoded:
I'm down but I think we'll have some model / endpoint changes coming. 
EX. having services have an affiliated dept. so we can pull related services via the dept. endpoint.
@benweatherman, based on convos with @cthibaultatx -- I think you'll have an info architecture/content relationship model WACKY AMAZING FUN shop thing coming up.

PR #57, opened 2018-01-10
Fix typo in README. Changed port from 800 to 8000 on line 25.

jessereitz:
Noticed the the port listed as 800 and figured it wasn't right. All good now!


benweatherman:
Thanks for the submission @jessereitz!

party

PR #59, opened 2018-01-12
Fix #34 : Actionable Title on ServicePage

### LINKED ISSUE #34: Use pythonic way to update help text for title
The current PR for updating the title helper text to say `Actionable Title` [is a JS hack](https://github.com/cityofaustin/joplin/pull/31#discussion-diff-157915453R3)

Instead it would be cool if this were a form field we could override. The problem is that `title` is part of the base `Page` model and seemed trickier to override.

jessereitz:
Fixes #34 | Creates an additional field for ServicePage, then sets the required/default title field to given input. It requires a default "Actionable Title" as well as a default slug, as the title is used to auto-generate the slug. The slug is automatically generated via the clean() method upon creation and validation of page. 

I'm not sure this is that much better than the JS hack which originally changed the label on the form. In fact, this may be worse/less stable in the long run given the default actionable title and slug values. That being said, it's the only way I can figure out without altering the core wagtail templates or models.


benweatherman:
Hey @jessereitz! Thanks for submitting this PR. Sorry it's taken so long to respond.

I think you're right in your assessment that the fix isn't better than the current JS hack. The only suggestion I have without digging into the problem more would be to look into overriding that field on the form https://github.com/cityofaustin/joplin/blob/5aac5ebe0ae18027f32369ae6e4c2d5702814c3f/joplin/base/forms.py#L4

That might allow us to override the field title without too much effort. Or maybe not 😼

jessereitz:
Hey @benweatherman! Right back at ya about the delay! haha. Yeah I'm thinking you're right about the override. I'll see if I can dig in and take a look but I'm thinking it may require us to make create a new db table with a one-to-one relationship to cover it. I'm not sure that'd be worth the time/effort/possible future confusion. I'm leaning more towards keeping the JS hack for the foreseeable future, eh?

benweatherman:
Agreed, keeping the JS hack is better than creating a new table. I really appreciate your research on this, even if we aren't able to merge the code.

I'll close this for now, but let's resurrect it if we can think of a better solution.

PR #60, opened 2018-01-15
Fix data in fixtures

benweatherman:
The fixture data was pointing to some incorrect keys. I added
instructions for making sure we don't run into this again by
using natural keys (which means the data isn't linked to a
specific PK but a model).


benweatherman:
Merging so everyone can use master again

PR #62, opened 2018-01-16
Update Chinese language codes

mateoclarke:


benweatherman:
Django uses the simplified and traditional chinese codes instead of the `CN` & `TW` codes. [Here's the bug](https://code.djangoproject.com/ticket/18419) they changed it in about 4 years ago. It seems like that's the standards-compliant way of doing things. It also seems like there's a lot of code out there that still uses `CN`/`TW`.

I think the way to fix this is to keep the hans/hant codes, but also accept `CN` as `hant` and `TW` as `hant`. That way we can support browsers that do non-standard things.

benweatherman:
Here are examples from master. Given this, I'm going to close this PR since Things Just Work™

## Django selects `zh-hans` when firefox and chrome send in `zh-cn`

#### Firefox
firefox - simplified

#### Chrome
chrome - simplified chinese

## Django selects `zh-hant` when firefox and chrome send in `zh-tw`

#### Firefox
firefox - traditional

#### Chrome
chrome - traditional chinese

PR #65, opened 2018-01-17
Stuff helpful for investigating translations

benweatherman:
In trying to understand how browsers use simplified and traditional
Chinese settings, I added some things that were helpful.

- Chinese, Spanish and Vietnamese data
- Translation panels within the CMS UI
- A django check to ensure languages are set correctly

Added a test script that will run all the system checks. Hopefully this
will one day be run automatically by some CI system as well as adding
real tests.

This is what we'd see when running `./scripts/test.sh` if someone, say, added `zh-cn` to the list of languages 😹 
tests


joplin/base/models.py
```diff
@@ -69,14 +69,28 @@ class ServicePage(Page):
 
     vi_panels = [
         # TODO: This field comes from Page and django-modeltranslation complains about it
-        # FieldPanel('title_es'),
+        # FieldPanel('title_vi'),
```
mateoclarke: ❓

benweatherman: This was a bad copypasta from the `es_panels` above. This comment keeps it consistent with the other panels by specifying the title field for this panel.

benweatherman:
The only change I _might_ make from that PR is reverting the codes to `zh-hans` and `zh-hant`. That seems like the more standardsy way of doing it. The weird thing is that chrome & firefox both send `zh-cn` and `zh-tw` so keeping them inline with browsers also seems like a good idea

https://github.com/cityofaustin/janis/pull/94/files#diff-684bb9ebdf7ae2c7c605b8341faf011eL19

joplin/base/models.py
mateoclarke: gotcha

PR #66, opened 2018-01-18
Service page images

mateoclarke:
I've added images to the ServicePage model using the example in the Wagtail docs.
http://docs.wagtail.io/en/v1.13.1/getting_started/tutorial.html?highlight=migration#images

Then I ran the makemigrations and migrate commands. 

When I try to run the local-serve command, the error I see is:

```python
django.db.utils.IntegrityError: Problem installing fixture '/app/fixtures/base.json': Could not load contenttypes.ContentType(pk=2): UNIQUE constraint failed: django_content_type.app_label, django_content_type.model
```

So, I ran the two commands related to dumping fixture data from the README.

Now I'm able to log in, and I see the Image field on a Service Page, but my fixture data is missing.

Halp 😨


mateoclarke:
closing this in favor of a fresh start

PR #81, opened 2018-01-26
Copy media folder to container

benweatherman:
Now that we're packing up images, we need to ship the `media` folder with the container so the preset images get served when deployed.

Fixes cityofaustin/joplin#70


benweatherman:
I'm gonna merge because it's such a simple one, but let me know if you see anything fishy

mateoclarke:
office_space_easy

PR #86, opened 2018-02-01
Allow any host in debug

benweatherman:
We don't care so much about the `ALLOWED_HOSTS` in debug mode and it only causes pain when trying to develop


benweatherman:
Merging because mateo said it works 😸

PR #88, opened 2018-02-01
Get ready for primetime

benweatherman:
`dj_database_url` parses the `DATABASE_URL` environment variable but we'll still fallback to sqlite. Conveniently, Heroku sets that env variable when we spin up a postgres instance. So hopefully we'll have a real DB when we're deployed.


benweatherman:
You should be able to login with the latest deployment. The fixture data loaded successfully.

data

benweatherman:
Deployed version is working, so I'm gonna merge this

PR #94, opened 2018-03-19
Custom fixtures

benweatherman:
This is an attempt at trying out a custom way of storing data until we settle on stuff in the CMS. Using just django fixture data made it hard to manage because of the sheer numbers of objects. Using YAML files hopefully makes it easy to see what data is getting loaded and hopefully it's an easy file to edit.

Examples of changes in the CMS that have caused 💔:
1. Migrating away from StreamFields The secondary body was initially added as a StreamField. We’ve now discovered that it’s not straightforward to translate those types of fields so we’re using a different tactic. Unfortunately it’s cumbersome to migrate this data automatically so I had to copypasta content to the new format
2. Adding translations to image alt text The builtin images don’t have the ability to store translations. That can be customized, but the images need to be re-uploaded and the alt text re-added.


benweatherman:
Merging so it's easier to implement on the Janis side of things

PR #96, opened 2018-03-21
Fixins

benweatherman:
- [x] Add departments
- [x] Images don’t load
- [x] Add contacts
- [x] Add dynamic content blocks
- [ ] Add hours for contacts (gonna do this in a separate PR so we can get the rest of this out there)


fixtures/contacts.yaml
```diff
@@ -0,0 +1,9 @@
+contacts:
+    -
+        name: 311
+        email: help@austintexas.gov
+        phone:
+            default: 311
```
ifsimicoded: data type change will require a Janis change in src/js/modules/ContactDetails and src/js/modules/ContactPhone

fixtures/locations.yaml
```diff
@@ -0,0 +1,8 @@
+locations:
+    -
+        name: Austin Recycle and Reuse Drop-off Center
```
ifsimicoded: I've been passing the contact name in -- 
we'll need to update the ContactDetails module to not pass in name to the ContactAddress module
and the ContactAddress module to use location.name
src/js/modules/ContactDetails and src/js/modules/ContactAddress

fixtures/services/bulk-item-pickup.yaml
```diff
@@ -111,6 +112,10 @@ additional_content_ar: |
         <li>السيارات المركونة.</li>
     </ul>
 
+dynamic_content:
+    -
+        type: collection_schedule
```
ifsimicoded: the type previously was collection_schedule_block.
We'll need to update the switch statement in src/js/modules/ApplicationBlock

fixtures/services/hazardous-waste-dropoff.yaml
```diff
@@ -54,6 +55,14 @@ additional_content_ar: |
     <p>لا تخلط العناصر. لن يتم قبول الأحمال الكبيرة للشاحنات أو المقطورة بعد الساعة 4:45 مساءً.</p>
     <p>الشركات ليست مؤهلة للتخلص من النفايات الخطرة. للحصول على معلومات حول كيفية تخلص الشركات من النفايات الخطرة، اتصل على الرقم 512-974-4336.</p>
 
+dynamic_content:
+    -
+        type: what_do_i_do_with
```
ifsimicoded: same as above -- We'll need to update the switch statement in src/js/modules/ApplicationBlock

fixtures/services/bulk-item-pickup.yaml
benweatherman: This won't change the API response

fixtures/services/hazardous-waste-dropoff.yaml
benweatherman: This won't change the API response

PR #103, opened 2018-03-22
Load phone number as JSON

benweatherman:
The phone number field is still text, but we want to serve it in the API as JSON. As a workaround, we'll just load the phone number as a JSON string. Someday we'll properly separate individual fields, but today ain't that day.

Part of cityofaustin/techstack#207

### DIFF
--- a/fixtures/contacts.yaml
+++ b/fixtures/contacts.yaml
@@ -3,9 +3,9 @@ contacts:
         name: 311
         email: help@austintexas.gov
         phone:
-            default: 311
-            out-of-area: 512-974-2000
-            tdd: 512-972-9848
+            default: 512-974-2000
+            n11: 311
+            tty: 512-972-9848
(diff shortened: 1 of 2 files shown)


fixtures/contacts.yaml
```diff
@@ -3,9 +3,9 @@ contacts:
         name: 311
         email: help@austintexas.gov
         phone:
-            default: 311
-            out-of-area: 512-974-2000
-            tdd: 512-972-9848
+            default: 512-974-2000
+            n11: 311
```
ifsimicoded: this might be a dumb, but is n11 specific to 311? will this key be changing for different contacts/phone numbers?

benweatherman: `n11` is just the term for 311, 911, etc. It sounded like we weren't going to display `311` so i removed it from default.

benweatherman:
Naw, I need to merge master into here to make it work

PR #107, opened 2018-03-24
Add tabbed recollect app to bulk pickup for cityofaustin/techstack#211

benweatherman:


benweatherman:
Got the 👌 from @cthibaultatx to merge this since it's not part of the user testing this round

PR #108, opened 2018-03-26
Fix Arabic translations for the trash & recycling pickup page

benweatherman:
These were still in English


benweatherman:
Got the 👌 from @cthibaultatx to merge since keeping it in English is unhelpful for Arabic user testing

PR #121, opened 2018-06-28
Lengthen fields so we can use more characters

benweatherman:
The deployed version of the app is failing when loading data

```
django.db.utils.DataError: value too long for type character varying(25)
```

We've got short titles that are at least 30 chars, so I'm just gonna make them long as funk so we don't have to think about this again.


fixtures/topics.yaml
```diff
@@ -437,7 +437,7 @@ topics:
     -
         slug: funding-grants-rebates
         theme: government-business
-        text_en: Fudning, grants, & rebates
+        text_en: Funding, grants, & rebates
```
mateoclarke: you can spell!

benweatherman: It was a requirement of joining the content team

PR #122, opened 2018-07-02
Delete images before loading them

benweatherman:
Django looks in the DB to see what images are available. We're running into a weird situation where the DB says there are images (the data is persisted because it's in postgres), but the images aren't on the filesystem (because we're using docker).

If we blow away the rows in the DB, django will remove the files on the filesystem if they exist. Then we can happily load the images and everything will be in harmony.

Other detritus:
- No need to register the delete image hooks since that bug has been fixed in wagtail
- Don't set loglevel to debug locally so we don't get so much cruft on the screen

Fixes cityofaustin/techstack#473

### DIFF
--- a/joplin/base/management/commands/loadcontent.py
+++ b/joplin/base/management/commands/loadcontent.py
@@ -7,10 +7,17 @@ from django.core.management.base import BaseCommand
 from wagtail.core.models import Page
 from yaml import load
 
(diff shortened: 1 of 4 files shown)


ifsimicoded:
Also, sorry I ruined your life, by introducing merge conflicts.

PR #124, opened 2018-07-09
Simplify UI to somewhat match mocks

benweatherman:
Fixes cityofaustin/techstack#302

### Bigger changes
- Hide everything in sidebar except for page and image listing
- Content list changes
  - Simplify header info
  - Add author column
  - Make edit/view/more buttons permanent (instead of only showing on hover)
- Edit content changes
  - Put action buttons at top instead of bottom
  - Remove red field separators
- Replace lovegoat with lovechicken

### Smaller changes
- Rename Promote panel to SEO
- Service page image (put it first on the page, update label)
- Bump minor wagtail version
- Add `django_extensions` so I could use `manage.py show_urls` to see which URLs are registered

### Result

content list

edit service page

image management


joplin/base/models.py
```diff
@@ -91,9 +89,7 @@ class ServicePage(Page):
 
     edit_handler = TabbedInterface([
         ObjectList(content_panels, heading='Content'),
-        ObjectList(Page.promote_panels, heading='Promote'),
-        # TODO: What should we do with the fields in settings?
-        # ObjectList(Page.settings_panels, heading='Settings', classname='settings'),
+        ObjectList(Page.promote_panels, heading='SEO'),
```
mateoclarke: Did we talk about doing this with @toribr? Just wondering bc of our repulsion to using confusing acronyms. "Metadata" or "Page Data" are some more accessible headers I'll throw out there for consideration.

joplin/base/models.py
```diff
@@ -65,24 +63,24 @@ class ServicePage(Page):
...
+        verbose_name='Add any maps or apps that will help the resident use the service',
     )
     additional_content = RichTextField(features=WYSIWYG_FEATURES, verbose_name='Write any additional content describing the service', blank=True)
     topic = models.ForeignKey(
         'base.Topic',
         on_delete=models.PROTECT,
         related_name='services',
     )
-    image = models.ForeignKey(TranslatedImage, null=True, blank=True, on_delete=models.SET_NULL, related_name='+')
+    image = models.ForeignKey(TranslatedImage, null=True, blank=True, on_delete=models.SET_NULL, related_name='+', verbose_name='Choose an image for the service banner')
```
mateoclarke: Should we give any suggestions here about aspect ratio or image sizing?

joplin/base/wagtail_hooks.py
```diff
@@ -47,6 +56,22 @@ def before_edit_page(request, page):
...
 
 
+@hooks.register('construct_main_menu')
+def configure_main_menu(request, menu_items):
+    new_items = []
+    for item in menu_items:
+        if item.name in ('home', 'images'):
+            item.label = ''
+            new_items.append(item)
+    menu_items[:] = new_items
```
mateoclarke: Ignorant question. What is this syntax? Not sure how to Google it.

joplin/templates/wagtailadmin/base.html
```diff
@@ -1,9 +1,19 @@
...
-    <img src="{% static 'images/logo.png' %}" alt="City of Austin" width="80" />
-    <script>
-        window.JANIS_URL = '{{ JANIS_URL }}';
-    </script>
+{% endblock %}
+
+{% block js %}
+    {{ block.super }}
+
+    <joplinheader>
```
mateoclarke: Whats up with this element naming? Would it make more sense for this to be `<div class="joplinheader">`

joplin/static/css/admin.css
```diff
@@ -1,9 +1,26 @@
...
-#tab-content .object .object-help {display: none;}
 .object fieldset, .object .field-content {width: 100%;}
 
 .translation-tab .object .object-help {width: auto; opacity: 1; position: static; font-size: 1.2em;}
 .translation-tab .object fieldset {padding-top: 0.5em;}
 
 .hidden {display: none !important;}
+
+.wrapper {padding-left: 60px;}
+.nav-wrapper, .nav-wrapper .inner {width: 60px;}
```
mateoclarke: There's something weird going on here when you scroll...
screen shot 2018-07-10 at 4 44 59 pm

joplin/base/models.py
benweatherman: We decided on SEO in our initial meeting https://docs.google.com/document/d/1znAEqCRBMJhBuyHJ3KGYUWtqFRTbtub5jMgCcg5qf1k/edit#heading=h.ymikv4aoyttz

> it would be nice to get the promote panel on the same page as in the mocks, but if it needs to be a separate tab/panel it will be labeled SEO 

We didn't talk through it further with content though

joplin/base/models.py
benweatherman: I used what's in the mockup. Adding help text for sizing makes sense. Should we make an issue?

It would be nice also if we could just make things work with aspect ratios or throw an error if it won't work, rather than typing out all the instructions.

joplin/base/wagtail_hooks.py
benweatherman: It's slice syntax. It's so we set the original `menu_items` objects instead of replacing it with a new list.

joplin/base/models.py
mateoclarke: Talked about this with Content Team and they like the label "Search Info"

benweatherman:
Thanks for the review @mateoclarke! I'll go ahead and merge this, but I'm definitely open to other changes or suggestions.

PR #126, opened 2018-07-12
Add revisions endpoint

briaguya:
This endpoint is to deliver draft data to Janis for use in previewing page content in Joplin.
closes https://github.com/cityofaustin/techstack/issues/479
related to https://github.com/cityofaustin/techstack/issues/478

### DIFF
--- a/joplin/api/schema.py
+++ b/joplin/api/schema.py
@@ -118,6 +118,11 @@ class ProcessPageNode(DjangoObjectType):
         interfaces = [graphene.Node]
 
 class PageRevisionNode(DjangoObjectType):
+    as_service_page = graphene.NonNull(ServicePageNode)
+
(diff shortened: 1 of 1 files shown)


mateoclarke:
I changed my language settingin browswer and checked in GraphiQL and the updated language came in the response. Requests seem good for intl.

screen shot 2018-07-13 at 3 18 15 pm

briaguya:
I also tested it out against my Janis branch and it gets the different languages when switching the slug

benweatherman:
Django won't serve assets unless `DEBUG=true` (which is set for staging and alpha). Hopefully this recent change will mean that the preview apps will also have it set https://github.com/cityofaustin/joplin/blob/40f4b67d840e5279b4cb39ca412f2adf09a50dcf/app.json#L11-L13

PR #127, opened 2018-07-12
Set page.title last so it's in english

benweatherman:
The last language we use is the language that'll show up by default in the wagtail admin. Set `page.title` last as the english version so it shows up in english.

english everywhere

### DIFF
--- a/joplin/base/management/commands/loadcontent.py
+++ b/joplin/base/management/commands/loadcontent.py
@@ -63,6 +63,7 @@ def load_theme(data):
 
     return theme
 
+
 def load_process(data):
 
     slug = data['slug']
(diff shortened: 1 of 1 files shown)


joplin/base/management/commands/loadcontent.py
```diff
@@ -72,7 +73,7 @@ def load_process(data):
     print('✅')
 
     print(f'-  Loading process steps...\r', end='')
-    process_steps = data.pop('process_steps');
+    process_steps = data.pop('process_steps')
```
ifsimicoded: you no like my decorative semi-colon? 😢

benweatherman: I'm sorry I removed your best contribution to the code

PR #128, opened 2018-07-14
Embed styleguide in edit page

benweatherman:
I widened the constants that were previously being used so the tabs could fit and more content could fit on the screen. Unfortunately this blocks some of the controls for editing so we might have to play with how this fits together. This seems like an OK start.

Fixes cityofaustin/techstack#467

Here's how things currently look:
styleguide

### DIFF
--- a/joplin/static/css/preview.css
+++ b/joplin/static/css/preview.css
@@ -1,14 +1,14 @@
 .preview-container {
   position: fixed;
-  right: 50px;
-  bottom: 50px;
+  right: 10px;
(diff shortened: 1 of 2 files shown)


benweatherman:
Totes

joplin/static/css/preview.css
```diff
@@ -1,14 +1,14 @@
 .preview-container {
```
briaguya: possibly rename this class, it contains more than just the preview

PR #129, opened 2018-07-14
Wire up Publish & Save Draft buttons

benweatherman:
Because the current button placement doesn't allow me to put them inside the form to submit, I hooked up the new Publish & Save Draft buttons to just click the old buttons from hidden bottom bar. I also hooked up the Preview button to do what the old preview button does, but we'll want to change that to hook into the new preview flow once we've got that nailed down.

Here's everything in action:
buttons

### DIFF
--- a/joplin/static/js/editor.js
+++ b/joplin/static/js/editor.js
@@ -1,0 +1,27 @@
+$(function() {
+
+  $('.js-proxy-click').click(function() {
+    let $this = $(this);
+    $this.text($this.data('clicked-text'));
+
+    let $button;
+
+    let proxyByName = $this.data('proxyByName');
+    if (proxyByName) {
+      $button = $(`[name="${proxyByName}"]`);
+    }
+
+    let proxyByClass = $this.data('proxyByClass');
+    if (proxyByClass) {
+      $button = $(`.${proxyByClass}`);
+    }
+
+    if (!$button) {
+      console.error(`Data attributes: ${$this.data()}`)
+      throw new Error('Unable to find a button. Did you specify data-proxy-by-name or data-proxy-by-class?');
+    }
+
+    $button.click();
+  });
+
+});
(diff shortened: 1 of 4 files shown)


joplin/static/js/editor.js
```diff
@@ -0,0 +1,27 @@
+$(function() {
+
+  $('.js-proxy-click').click(function() {
+    let $this = $(this);
+    $this.text($this.data('clicked-text'));
+
+    let $button;
```
mateoclarke: We'll need to get babel in here if we're going to use fancy js. `let` doesn't work before IE11 and I expect we'll have some editors using shitty versions of IE. https://caniuse.com/#feat=let

joplin/static/js/editor.js
```diff
@@ -0,0 +1,27 @@
...
+    if (proxyByClass) {
+      $button = $(`.${proxyByClass}`);
+    }
+
+    if (!$button) {
+      console.error(`Data attributes: ${$this.data()}`)
+      throw new Error('Unable to find a button. Did you specify data-proxy-by-name or data-proxy-by-class?');
+    }
+
+    $button.click();
```
mateoclarke: Why did we have to do this proxy click dance instead of letting the new buttons do the submit actions? Guessing it has to do with limitations to the existing html layouts? This seems a little hacky, but I'm not above it.

joplin/static/js/editor.js
benweatherman: Ah, forgot about that. I can go oldskool.

joplin/static/js/editor.js
benweatherman: I thought it might've saved us some headaches if they were doing anything weird in a JS click handler, but it's probably easier for future readers if I just submit the form. I can give that a whirl and see where it lands.

briaguya: I have them wired up to hit submit in my PR: https://github.com/cityofaustin/joplin/pull/132

benweatherman: OK Sweet, so it seems like this PR isn't even necessary? I was just trying to wire something up quickly to kinda see things work, but @briaguya has the real deal going on.

mateoclarke:
Based on the convo about Brian implementing this in another PR #132, lets close this without merging for now.

PR #130, opened 2018-07-14
Kickoff Heroku build when pages are published/unpublished

benweatherman:
I've noticed it takes about 10 minutes for the new build to finish. I started
to add the ability to show the build progress via the UI, but wasn't sure what
that workflow should look like.

Fixes cityofaustin/techstack#480

### DIFF
--- a/joplin/settings.py
+++ b/joplin/settings.py
@@ -182,7 +182,8 @@ SECRET_KEY = '1iq1u6gs+xh3!bvrl-5$jqne%gpj)!wv5^h0$dc0y84xsdr-95'
 
 EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 
-
+HEROKU_KEY = os.getenv('HEROKU_KEY')
+HEROKU_JANIS_APP_NAME = os.getenv('HEROKU_JANIS_APP_NAME')
 JANIS_URL = os.getenv('JANIS_URL', 'http://localhost:3000')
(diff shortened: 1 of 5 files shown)


joplin/settings.py
```diff
@@ -182,7 +182,8 @@
 
 EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 
-
+HEROKU_KEY = os.getenv('HEROKU_KEY')
+HEROKU_JANIS_APP_NAME = os.getenv('HEROKU_JANIS_APP_NAME')
```
mateoclarke: So I think I understand how we would get janis-staging or austindotgov heroku apps to rebuild upon a page publish from local based on your instructions.

But in order for this to work not just from local, but from a (un)publish event in `joplin-staging` or `joplin` we already have the env vars we need set, right?

benweatherman: I can't login to heroku, but I think I added the vars as stubs. To get it fully setup, I think you'll need to create a new user and add a non-expiring token for them. The [auth docs](https://devcenter.heroku.com/articles/platform-api-quickstart#authentication) say

> You can create a non-expiring token by running heroku authorizations:create

Then you'll need to add that token to the config vars section of the app(s).

mateoclarke:
we should be good to merge this but lets test it early in the day/week so we dont leave something broken over the weekend.

PR #134, opened 2018-07-24
Style guide nav links

briaguya:
Update the edit page field labels to link to anchors in the style guide in the sidebar


briaguya:
jul-24-2018 04-06-00

joplin/static/js/admin.js
```diff
@@ -1,12 +1,55 @@
-document.addEventListener('DOMContentLoaded', function(event) {
+// TODO: This a better way
+const anchorLinks = {
+  id_title:
```
mateoclarke: what if we had a key that was the `baseurl` so then each of the following `id_foo` are much shorter? ex:

```
{
  basurl:  `https://briaguya.github.io/digital-services-style-guide/writing-service-pages/`,
  id_steps: `#steps`,
  ..
}
```

joplin/static/js/admin.js
```diff
@@ -1,12 +1,55 @@
...
+          break;
+      }
+      debugger;
+    }
+
+    if (id in anchorLinks) {
+      text = document.createTextNode(label.textContent);
+      label.textContent = "";
+      if (id === "id_title") {
+        text = document.createTextNode("Actionable Title");
```
mateoclarke: This if statement is a little smelly. Why is there this one-off?

joplin/static/js/admin.js
```diff
@@ -1,12 +1,55 @@
...
+          break;
+        case "CONTACTS":
+          id = "id_contacts";
+          break;
+      }
+      debugger;
+    }
+
+    if (id in anchorLinks) {
+      text = document.createTextNode(label.textContent);
```
mateoclarke: don't we need to use keyword `var` before text? And label & link below?

joplin/static/js/admin.js
```diff
@@ -1,12 +1,55 @@
...
+    }
   }
 
   // HACK: I can't find a way to configure this to be closed via python
-  document.querySelectorAll('#extra_content-prependmenu:not(.stream-menu-closed)').forEach((elem) => {
-    elem.click();
-  });
+  document
+    .querySelectorAll("#extra_content-prependmenu:not(.stream-menu-closed)")
+    .forEach(elem => {
```
mateoclarke: We need Babel yesterday if we expect to write this fancy js for gov employees on crap Windows boxes.
https://caniuse.com/#feat=arrow-functions

briaguya: agreed

joplin/static/js/admin.js
briaguya: This was a one off hack that was already in the codebase:
```
// HACK: I can't find a way to override this in python
const title = document.querySelector('label[for=id_title]');
if (title) {
  const labels = document.querySelectorAll("label");
  title.textContent = 'Actionable Title';
```

joplin/api/schema.py
```diff
@@ -119,10 +119,14 @@ class Meta:
 
 class PageRevisionNode(DjangoObjectType):
     as_service_page = graphene.NonNull(ServicePageNode)
+    as_process_page = graphene.NonNull(ProcessPageNode)
 
     def resolve_as_service_page(self, resolve_info, *args, **kwargs):
         return self.as_page_object();
 
+    def resolve_as_process_page(self, resolve_info, *args, **kwargs):
```
mateoclarke: This seems like it might get unwieldy overtime. Is there a path to refactor this towards `resolve_as_page` and get the page type via attributes?

joplin/static/css/preview.css
```diff
@@ -8,11 +8,10 @@
 /* This container helps the thumbnail behave as if it were an unscaled IMG element */
 .thumbnail-container {
   width: 100%;
-  display: inline-block;
+  /*display: inline-block;*/
```
mateoclarke: lets just delete this line

joplin/static/css/preview.css
```diff
@@ -44,7 +43,7 @@
 .thumbnail,
 .thumbnail iframe {
   width: 100%;
-  height: 75vh;
+  height: 72vh;
```
mateoclarke: just highlighting this as a place where it would be nice to have Sass variables.

joplin/static/css/preview.css
```diff
@@ -74,6 +73,22 @@
...
+  display: flex;
+}
+
+.sidebar-button-container {
+  margin: auto;
+}
+
 .fixed-content {
   position: fixed;
+  box-shadow: 0 0 2px 2px #666;
```
mateoclarke: 👹6️⃣6️⃣6️⃣👹

joplin/static/js/admin.js
```diff
@@ -1,12 +1,55 @@
-document.addEventListener('DOMContentLoaded', function(event) {
+// TODO: This a better way
+const anchors = {
```
mateoclarke: 👍

scripts/deploy.sh
```diff
@@ -10,4 +10,8 @@ if [[ $CURRENT_CONFIG != *"JANIS_URL"* ]]; then
     heroku config:set LOAD_DATA=on JANIS_URL=https://janis-staging.herokuapp.com --app "$APP"
 fi
 
+if [[ $CURRENT_CONFIG != *"STYLEGUIDE_URL"* ]]; then
```
mateoclarke: 💅

scripts/serve-local.sh
```diff
@@ -31,4 +31,5 @@ docker run \
     --env "LOAD_DATA=$LOAD_DATA" \
     --env "GUNICORN_CMD_ARGS=--reload  --reload-engine=poll" \
     --env "JANIS_URL=http://localhost:3000" \
+    --env "STYLEGUIDE_URL=https://cityofaustin.github.io/digital-services-style-guide" \
```
mateoclarke: can you make sure to add this env var to heroku if you haven't already so we don't forget and break the build?

mateoclarke:
@briaguya I'm getting this error locally on this branch when I try to edit a service or process page.

screen shot 2018-08-02 at 11 48 39 am

mateoclarke:
@briaguya forget my last comment, restarting fixed it. I think I might've started my server in master and switched branches so that env var wasn't available from startup. My b

joplin/api/schema.py
briaguya: I would love to make that possible, however, as_page_object() returns either a ServicePageNode or a ProcessPageNode. I tried abstracting it and using a PageNode before but I ran into a type error.

Overall I'm not super concerned about it, we already are making different graphQL queries based on the different properties we want for any given page.

PR #139, opened 2018-08-03
Bs edit page improvements

briaguya:
https://github.com/cityofaustin/techstack/issues/533
https://github.com/cityofaustin/techstack/issues/571

### DIFF
--- a/joplin/static/css/preview.css
+++ b/joplin/static/css/preview.css
@@ -42,7 +42,7 @@
 .thumbnail,
 .thumbnail iframe {
   width: 100%;
-  height: 72vh;
+  height: calc(100vh - 242px);
 }
 /* This facilitates the fade-in transition instead of flicker. It also helps us maintain the illusion that this is an image, since some webpages will have a preloading animation or wait for some images to download */
 .thumbnail iframe {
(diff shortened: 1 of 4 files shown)


joplin/static/css/preview.css
```diff
@@ -42,7 +42,7 @@
 .thumbnail,
 .thumbnail iframe {
   width: 100%;
-  height: 72vh;
+  height: calc(100vh - 242px);
```
mateoclarke: 👏

mateoclarke: just double checked caniuse.com and we can use `calc` back to IE9 with partial support. Wondering if we should get some usage stats or set a support standard for the CMS. We went through that process with the resident site, but the data would be different for city admins than residents. Probably for the worse 😞

joplin/static/css/preview.css
```diff
@@ -64,16 +64,19 @@
 /* Improving the sidebar experience with flexbox */
 .flex-container {
   display: flex;
+  margin-top: 204px;
```
mateoclarke: lots of magic numbers here are another reason we need Sass yesterday

joplin/static/js/admin.js
```diff
@@ -12,20 +12,24 @@ const anchors = {
   "id_process_steps-description": "#step-description",
   "id_process_steps-overview_steps": "#step-steps",
   "id_process_steps-detailed_content": "#step-details",
-  "id_process_steps-quote": "#step-quote",
+  "id_process_steps-quote": "#step-quote"
```
mateoclarke: Same thing I said in the other PR, but I think we should be wrapping our JS in either a document.ready or IIFE so that we don't get burned by global scopes or eager browsers.

PR #140, opened 2018-08-03
Bs share button improvements

briaguya:
https://github.com/cityofaustin/techstack/issues/542


briaguya:
comments will be addressed in https://github.com/cityofaustin/techstack/issues/592

PR #141, opened 2018-08-09
New Content Modal in React

mateoclarke:
Closes https://github.com/cityofaustin/techstack/issues/415

aug-22-2018 16-34-43


joplin/css/create_content_modal.scss
```diff
@@ -0,0 +1,251 @@
...
+.ChooseTopicStep__topic--selected {
+  background-image: url('../static/images/selected.png');
+  height: 22px;
+  width: 20px;
+  margin-left: 1px;
+  margin-right: 13px;
+}
+
+
+// BUTTON BAR COMPONENT
```
briaguya: might be nice to move these to their own files so you can just include the buttonbar styles in the buttonbar component etc

joplin/js/CreateContentModal/ButtonBar.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+
+const buttonRowClassName = (activeStep) => {
```
briaguya: maybe we could use [classnames](https://github.com/JedWatson/classnames) for this instead?

joplin/js/CreateContentModal/ChooseTopicStep.js
```diff
@@ -0,0 +1,82 @@
...
+
+const TopicItem = ({ topic, handleTopicSelect, activeTopic }) => (
+  <li
+    onClick={handleTopicSelect.bind(this, topic.id)}
+    className="ChooseTopicStep__topic-toggle"
+  >
+    <div
+      className={
+        activeTopic === topic.id
+          ? "ChooseTopicStep__topic--selected"
```
briaguya: Maybe this could be fancier with:
``` `ChooseTopicStep__topic--${activeTopic === topic.id ? "selected" : "unselected"}` ```

after writing this not sure which way is more readable though

File: joplin/js/CreateContentModal/ChooseTopicStep.js [hunk 4]
```diff
@@ -0,0 +1,82 @@
...
+    };
+  }
+
+  handleThemeToggle = (id, e) => {
+    // This collapses the currently selected Theme Group when it is clicked
+    if (Number(id) === this.state.openThemeGroup) {
+      this.setState({
+        openThemeGroup: 0,
+      });
+      return true;
```
briaguya: where are we using this return value?

joplin/js/CreateContentModal/ChooseTypeStep.js
```diff
@@ -0,0 +1,53 @@
...
+        >
+          <img
+            src={processPageImage}
+            alt="Service Page"
+          />
+          <h3 className="ChooseTypeStep__option-header">Process Page</h3>
+          <p>Processes which require several steps, which may not go in order</p>
+        </div>
+        <div className="ChooseTypeStep__option"
+          onClick={handleTypeSelect.bind(this, {
```
briaguya: I think `onClick={() => handleTypeSelect({YOUR_DATA_HERE})}` should work here

joplin/js/CreateContentModal/index.js
```diff
@@ -0,0 +1,133 @@
...
+              </div>
+            </div>
+          </div>
+        </div>
+      </div>
+    );
+  }
+}
+
+ReactDOM.render(
```
briaguya: Do we want the ReactDOM.render here? I kind of like the pattern set by create-react-app where you have App.js with your app and then index.js that just imports it and renders it. Not that we need to stick to that naming convention it's just nice to separate the component from the rendering.

joplin/static/js/admin.js
```diff
@@ -0,0 +1,68 @@
+// TODO: This a better way
```
briaguya: I don't think this file is supposed to be here since we refactored to have static just include the bundles

joplin/templates/wagtailadmin/pages/create.html
```diff
@@ -62,42 +62,4 @@ <h1 class="icon icon-doc-empty-inverse">
 
 {% block extra_js %}
```
briaguya: if this is just rendering the super block then we can delete this file

joplin/webpack.build.js
```diff
@@ -1,38 +1,6 @@
-var BundleTracker = require("webpack-bundle-tracker");
-var path = require("path");
+var merge = require('webpack-merge');
```
briaguya: ooh! this looks cleaner, I'm excited to see how it works!

joplin/css/create_content_modal.scss
mateoclarke: ✅

joplin/js/CreateContentModal/ButtonBar.js
mateoclarke: ✅

joplin/js/CreateContentModal/ChooseTopicStep.js
mateoclarke: because if we don't return it continues on to line 38. Do you mean, we should just `return` with no value?

joplin/js/CreateContentModal/ChooseTypeStep.js
mateoclarke: ✅

joplin/js/CreateContentModal/index.js
mateoclarke: I don't have a strong opinion about this. Having an extra file that just renders it doesn't seem useful to me but if we wanted to add fancy things like hot module reloading, I think it would become necessary.

PR #143, opened 2018-08-14
Webpack 2: Electric Boogaloo

briaguya:


Dockerfile
```diff
@@ -1,5 +1,10 @@
 FROM python:3.6.5-slim-stretch
```
briaguya: I was really hoping to be able to do ```FROM joplin-base``` here but I couldn't figure out how to get Heroku to build the base dockerfile into an image before trying to build this one, so for now there's a bit of copypasta.

I'd like to put this off into an issue and tag it as techdebt if y'all are cool with that instead of doing more digging right now.

Dockerfile
```diff
@@ -12,8 +17,13 @@ RUN mkdir /app
 WORKDIR /app
 
 COPY "$PWD/fixtures" /app/fixtures
-COPY "$PWD/docker-entrypoint.sh" /app/docker-entrypoint.sh
 COPY "$PWD/joplin" /app/joplin
 
-ENTRYPOINT ["./docker-entrypoint.sh"]
+COPY "$PWD/migrate-load-data.sh" /app/migrate-load-data.sh
+RUN LOAD_DATA=on ./migrate-load-data.sh
```
briaguya: By doing this as a RUN command instead of having it in the ENTRYPOINT, it means that when we boot up the built image on heroku it only boots up wagtail instead of migrating and loading data.

Dockerfile.local
```diff
@@ -0,0 +1,4 @@
+FROM joplin-base
+
+ENTRYPOINT ["./docker-entrypoint.sh"]
```
briaguya: By keeping this entrypoint in our local Dockerfile, we can still use our LOAD_DATA command without needing to rebuild the image

joplin/css/admin.scss
```diff
@@ -0,0 +1,78 @@
+.full input {
```
briaguya: This is pretty much just a rename of our admin.css, but since it's been moved and renamed GH seems to think it's new

joplin/js/admin.js
```diff
@@ -0,0 +1,10 @@
+import "../css/admin.scss";
```
briaguya: same stuff we saw with admin.css (except I moved a ton of the editor/preview logic out to their respective js files)

joplin/webpack.build.js
```diff
@@ -0,0 +1,38 @@
+var BundleTracker = require("webpack-bundle-tracker");
```
briaguya: Little bit of copypasta between this and webpack.dev.js, any ideas on how to reduce that are super welcome

scripts/serve-local.sh
```diff
@@ -2,8 +2,6 @@
 
 set -o errexit
 
-TAG='joplin:local'
```
briaguya: Not tagging because we're using docker-compose

Dockerfile
mateoclarke: +1 to bookmarking it as devops tech debt in a gh-issue

Dockerfile
```diff
@@ -1,5 +1,10 @@
 FROM python:3.6.5-slim-stretch
 
+RUN apt-get update; apt-get -y install gnupg; apt-get -y install curl
+RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
+RUN apt-get update; apt-get -y install nodejs
```
mateoclarke: what version of nodejs are we running here?

Dockerfile
mateoclarke: So does this mean that we were previously blowing away prod data each deploy and we won't be doing that anymore?

docker-compose.local.yml
```diff
@@ -0,0 +1,24 @@
+version: '3'
```
mateoclarke: Is there or will there need to be a docker-compose yml for other envs? Or are we not needing docker-compose for staging and prod?

docker-entrypoint.sh
```diff
@@ -1,24 +1,6 @@
...
-        fixtures/311.yaml \
-        fixtures/themes.yaml \
-        fixtures/topics.yaml \
-        fixtures/locations.yaml \
-        fixtures/contacts.yaml \
-        fixtures/departments.yaml \
-        fixtures/services \
-        fixtures/processes
-fi
+./migrate-load-data.sh
```
mateoclarke: this seems good 💅

joplin/base/wagtail_hooks.py
```diff
@@ -10,46 +10,6 @@
 from base.models import HomePage, Topic, Location, Contact
 
 
-@hooks.register('insert_global_admin_css')
```
mateoclarke: So we're removing the asset inserting here in the py files and the new code is what happens with `{% render_bundle 'bundle_name' %}` in the html files?

It seems like those bundles right now get collected at the entrypoints. Do those have to be js files? What if all the customization we are doing to a template is in the css? I think I'm missing a step here in my understanding

joplin/settings.py
```diff
@@ -151,6 +152,13 @@
     os.path.join(PROJECT_DIR, 'static'),
 ]
 
+WEBPACK_LOADER = {
+    'DEFAULT': {
+        'BUNDLE_DIR_NAME': 'webpack_bundles/',
+        'STATS_FILE': os.path.join(BASE_DIR, 'joplin/static/webpack-stats.json'),
```
briaguya: Make a note in the readme telling people to delete their webpack bundles folder if it gets too intense (also maybe make a techdebt issue to make this not a problem ever)

joplin/webpack.build.js
briaguya: let's make a techdebt issue

scripts/serve-local.sh
```diff
@@ -24,18 +22,10 @@ fi
...
-    --volume "$PWD:/app" \
-    --env "DEBUG=1" \
-    --env "LOAD_DATA=$LOAD_DATA" \
-    --env "GUNICORN_CMD_ARGS=--reload  --reload-engine=poll" \
-    --env "HEROKU_KEY=$HEROKU_KEY" \
-    --env "HEROKU_JANIS_APP_NAME=$HEROKU_JANIS_APP_NAME" \
-    --env "JANIS_URL=http://localhost:3000" \
-    --env "STYLEGUIDE_URL=https://cityofaustin.github.io/digital-services-style-guide" \
-    "$TAG" "$@"
+if [ "$REBUILD" == "on" ]; then
```
briaguya: update readme to discuss rebuild

Dockerfile
briaguya: https://github.com/cityofaustin/techstack/issues/604

Dockerfile
briaguya: 10.x

docker-compose.local.yml
briaguya: Currently we're doing all the build webpack work that's happening in the assets dockerfile as part of the build in the heroku Dockerfile. If this becomes a problem we can investigate other solutions. I don't think it's worth making a techdebt issue for yet.

joplin/base/wagtail_hooks.py
briaguya: For now I think the best path would be to create a .js file that just imports the .scss files.

joplin/webpack.build.js
briaguya: https://github.com/cityofaustin/techstack/issues/605

PR #153, opened 2018-08-21
seo fields in main content block and custom titles

ifsimicoded:
This branch demonstrates how to update wagtail to 
1) have a verbose name for a default model field (title) and custom model field (topic)
screen shot 2018-08-21 at 4 02 20 pm

2) have the SEO fields in the same panel as the primary content.
seo_panel

It might make sense to extend the base Page class and build our models of the extended class so we don't have to do this for every page type. Something like ... but I don't know python well to say this is the right syntax

```
class COAPage(Page):
   created_at = models.DateTimeField(auto_now_add=True)
   updated_at = models.DateTimeField(auto_now=True)
   
   parent_page_types = ['base.HomePage']

   seo_panels = (
        FieldPanel('slug'),
        FieldPanel('seo_title'),
        FieldPanel('show_in_menus'),
        FieldPanel('search_description'),
    )
    content_panels = [ 
        MultiFieldPanel(seo_panels, heading='Search Engine Optimization', classname='seo_panels')
    ]
    ...

COAPage._meta.get_field('title').verbose_name='Write an actionable title'

class MyPage(COAPage):
    content_panels = super.content_panels + [ 
      ...
    ]

```

### DIFF
--- a/joplin/base/models.py
+++ b/joplin/base/models.py
@@ -55,6 +55,13 @@ class ServicePage(Page):
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
 
+    seo_panels = (
+        FieldPanel('slug'),
+        FieldPanel('seo_title'),
+        FieldPanel('show_in_menus'),
+        FieldPanel('search_description'),
+    )
+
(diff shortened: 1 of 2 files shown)


mateoclarke:
I'm going to close this out. This might be valuable as reference but its linked in the related issue about adding back the SEO tab.

PR #154, opened 2018-08-21
Update locations.yaml

LauraTrujillo:
Added 2 public health clinic addresses


fixtures/locations.yaml
```diff
@@ -6,3 +6,17 @@ locations:
         state: TX
         zip: 78744
         country: United States
+
```
mateoclarke: you need a dash here on line 9 and on line 16 the way there is one on line 2

PR #156, opened 2018-08-22
fix public health yaml

ifsimicoded:
Images and Locations still need to be inserted and updated.
Based on this content, steps will need to be modified to:
- include a title
- be nest-able?

@LauraTrujillo I'm of the mindset that the content should be restructured to not require so much nesting -- but I'm not sure. Maybe we should discuss?
@briaguya I worry that our concurrent changes with steps (which don't comply with what the public health content requires) will leave us with broken public health pages -- or having to redo this yaml. 
Also, should yarn.lock be checked in?


ifsimicoded:
hey @briaguya @LauraTrujillo, thanks for the review! There's some minor html glitches in the new yaml that got left behind, but I can open a new branch to fix those. Also @briaguya, I'm not sure what's going on with react-static -- but clicking a link to a page shows the content appropriately -- navigating to a page directly (or refreshing a page) shows the content missing sub-steps. @briaguya I'll log a new issue for that one. Closing this PR out.

PR #157, opened 2018-08-22
Bs fix live links

briaguya:
this covers:
https://github.com/cityofaustin/techstack/issues/614
https://github.com/cityofaustin/techstack/issues/578


joplin/base/models.py
```diff
@@ -52,9 +54,26 @@ class HomePage(Page):
     image = models.ForeignKey(TranslatedImage, null=True, on_delete=models.SET_NULL, related_name='+')
 
 
-class ServicePage(Page):
+class JanisPage:
```
ifsimicoded: 🙏

briaguya: @ifsimicoded re: extending - I remember running into an issue when trying to do it that way, don't remember what it was. I'll go ahead and try it out in a new branch to see if I can figure out what it was.

briaguya: just checked and remember what it was I ran into:
```
joplin_1        |     return Database.Cursor.execute(self, query, params)
joplin_1        | sqlite3.OperationalError: no such table: base_basejanispage
joplin_1        |
```

I didn't go through all the hoops to use migrations update our models to include a BaseJanisPage to the DB which would be needed to do the inheritance for the model instead.

This should be addressed by https://github.com/cityofaustin/techstack/issues/634

PR #160, opened 2018-08-27
Data & File Persistance

sergiogcx:
This is a test for me to understand the current deployment process, I have made a small change in the start-local.sh file and created a branch which I am trying to run on my own PR instance.


sergiogcx:
As of this moment, data is persisted. The only problem is that the migration process has to be run manually (it will not kick off automatically on Heroku). 

Images will need to be persisted too, as soon as the app instance is terminated, any new image uploads will be gone too. These images will have to be uploaded to an S3 bucket.

docker-compose.local.yml
```diff
@@ -1,20 +1,33 @@
 version: '3'
 services:
+  joplindb:
+    image: postgres
+    ports:
+     - "5432:5432"
+    environment:
+      POSTGRES_DB: 'joplin'
+      POSTGRES_USER: 'joplin'
+      POSTGRES_PASSWORD: '39h3zd3j9Pc57mH3QsiS6sb4'
```
mateoclarke: Is this ok to have captured in Git?

docker-compose.local.yml
```diff
@@ -1,20 +1,33 @@
...
+     - ".:/app"
     environment:
       LOAD_DATA: $LOAD_DATA
       HEROKU_KEY: $HEROKU_KEY
       DEBUG: 1
       GUNICORN_CMD_ARGS: "--reload  --reload-engine=poll"
       JANIS_URL: http://localhost:3000
       STYLEGUIDE_URL: https://cityofaustin.github.io/digital-services-style-guide
+      DEPLOYMENT_MODE: "LOCAL"
+      DATABASE_URL: "postgres://joplin:39h3zd3j9Pc57mH3QsiS6sb4@localhost:5432/joplin"
```
mateoclarke: Does setup on local change for us using postgres now? If so, is there any documentation or instructions we should add to the README for setup?

docker-entrypoint-prod.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+#echo "Running migrations (Database & static files to S3 bucket) ..."
+echo "Bucket: ${AWS_S3_BUCKET}"
+echo "Bucket User: ${AWS_S3_USER}"
+echo "Bucket ID: ${AWS_S3_KEYID}"...
+
+
+function send_message {
+    curl -s \
+      --form-string "token=avu1jex2ixynqx3waz92hwrjqdezrt" \
```
mateoclarke: Brian just explained to me that we only plan to use these tokens in the short run until we figure out how to load secret env vars into our Docker containers with some of the Heroku constraints. carry on...

docker-entrypoint-prod.sh
```diff
@@ -0,0 +1,28 @@
...
+      --form-string "user=ufuvXYqcU6fnJbRCh7Uc9qktcbARuv" \
+      --form-string "message=$1" \
+      https://api.pushover.net/1/messages.json
```
mateoclarke: what is the Pushover API do?

joplin/settings.py
```diff
@@ -139,6 +148,9 @@
 TIME_ZONE = 'UTC'
 USE_TZ = True
 
+MODELTRANSLATION_DEFAULT_LANGUAGE = 'en'
```
mateoclarke: 🔤

docker-compose.local.yml
sergiogcx: That should be fine in my opinion since it is for local use only and does not affect any PR, staging or production deployments. It may be possible to set up a no-password requirement; not sure how long this might take.

docker-entrypoint-prod.sh
sergiogcx: Ah, darn yes let me get rid of it. I was trying to test a Heroku Procfile and I could not determine if it was running or not, hence the pushover. This function needs to be deleted.

docker-entrypoint-prod.sh
sergiogcx: We are not using tokens, they work fine in Heroku and the migration process is responding well. The only challenge is that the migration process can take very long, and Heroku does have a 60 seconds boot timeout limit. If the service isn't live within 60 seconds, then it will take the app down.

joplin/settings.py
sergiogcx: In my local environment, the migration process inserted the data without errors; however, when inserting a multi-language entry it kept defaulting to Arabic; for example, when it inserted a page, the "title_en", "title_e"s, "title_ar", and "title_vi" columns inserted the content fine, but the "title" column defaulted to title_ar.

I tried manually assigning the values title = title_en, but even that would not work it would re-write the entire title_ar column with it. When inserted this line, the problem went away.

docker-compose.local.yml
sergiogcx: This is for local use only, it should not be a security issue since it is not run on the cloud. This could probably be changed so that postgres does not require a password. I will have to play around with the settings.

sergiogcx:
My only concern at this point is that Heroku has a 60 second boot timeout. If the migration process takes longer it will shut down and take down the service. It could be extended to 120 seconds, but I do not think we should count with this time to make the deployment.

I am trying to work out a way to run an independent Dyno process that can run the migration script independently.

docker-entrypoint-prod.sh
mateoclarke: should or can we make the timeout longer?

joplin/settings.py
mateoclarke: yeah, I remember you explaining that. Fixed in just one line 👍

docker-compose.local.yml
briaguya: it might make sense to change this password to reflect the fact that it's for local use and insecure, something like `POSTGRES_PASSWORD: insecureLocalOnlyPassword#123` instead of having it look like a somewhat secure random string of numbers chillin in the codebase

docker-compose.local.yml
briaguya: re: @mateoclarke's comment, I don't expect any of the steps in the readme to change beyond what has changed in the move to docker-compose from using a single dockerfile. If there's something more complex than that (especially regarding rebuilding/wiping the db) then we absolutely should add it to the readme

docker-entrypoint-prod.sh
briaguya: @mateoclarke we can increase the timeout but I don't think that's the best method to address this issue. The timeout is for booting up an already built container, so running migrations during that boot process is less than ideal. I addressed this issue earlier when we started getting timeouts when populating data from our yaml files in the entrypoint scripts.

The real solution should be to figure out how to use environment variables or other secure methods to store AWS keys etc. within the docker build process. That way we aren't trying to run it every time we boot up the container and we can keep our restart on failure times as short as possible.

docker-compose.local.yml
sergiogcx: Agreed, the steps remain the same. @briaguya @mateoclarke

docker-compose.local.yml
sergiogcx: It looks like it works fine without any password set up, tested it repeatedly via redeployment and with two separate clients. Will commit change to branch.

.travis.yml
```diff
@@ -21,7 +21,9 @@ before_install:
     - source ./.travis/heroku-helper.sh
     - sudo pip install awscli
     - curl https://cli-assets.heroku.com/install-ubuntu.sh | sh
-    - sudo apt-get purge -y postgresql && sudo apt-get install postgresql-10 postgresql-client-10
+    - wget -q https://www.postgresql.org/media/keys/ACCC4CF8.asc -O - | sudo apt-key add -
+    - sudo apt-get update && sudo apt-get install postgresql postgresql-contrib
+    #- sudo apt-get purge -y postgresql && sudo apt-get install postgresql-10 postgresql-client-10
```
briaguya: probably a good idea to remove this commented section unless we plan on going back to it later

PR #162, opened 2018-08-29
Content Index nav and header updates

mateoclarke:
Closes https://github.com/cityofaustin/techstack/issues/613

~We should update alternative icons for left sidebar so they swap colors depending on active vs. inactive state. I've asked Tori for updated assets (SVG that we can change the `stroke` value or PNGs for each color).~

screen shot 2018-08-29 at 5 51 04 pm


joplin/css/admin.scss
```diff
@@ -58,11 +62,11 @@
 }
 .header {
   display: flex;
-  padding: 0.75rem;
-  background: #333;
+  padding: 0.75rem 1.5rem;
+  background: $coa-color-darkest-gray;
```
briaguya: 😃

PR #166, opened 2018-09-07
Base Page Model

ifsimicoded:
Closes https://github.com/cityofaustin/techstack/issues/634

This is a starting point for a shared base model.
@briaguya @mateoclarke -- I'm not 100% on how inheritance works in Django so feedback plz. 

Required related changes in Janis here: https://github.com/cityofaustin/janis/pull/268

### DIFF
--- a/joplin/api/schema.py
+++ b/joplin/api/schema.py
@@ -104,7 +104,7 @@ class ServicePageNode(DjangoObjectType):
     related = graphene.List('api.schema.ServicePageNode')
 
     def resolve_related(self, resolve_info, *args, **kwargs):
-        return self.topic.services.exclude(id=self.id)
+        return self.topic.servicepage_set.exclude(id=self.id)
 
     class Meta:
         model = ServicePage
--- a/joplin/base/models.py
+++ b/joplin/base/models.py
@@ -5,6 +5,7 @@ import graphene
 from modelcluster.fields import ParentalKey
 from modelcluster.models import ClusterableModel
 
+from wagtail.utils.decorators import cached_classmethod
 from wagtail.admin.edit_handlers import FieldPanel, InlinePanel, MultiFieldPanel, ObjectList, StreamFieldPanel, TabbedInterface
 from wagtail.core.fields import StreamField, RichTextField
 from wagtail.core.models import Page, Orderable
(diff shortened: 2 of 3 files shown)


joplin/base/models.py
```diff
@@ -55,7 +56,39 @@ class HomePage(Page):
...
+        'base.Topic',
+        on_delete=models.PROTECT,
+        verbose_name='Select a Topic',
+    )
+    content_panels = [
+        FieldPanel('topic'),
+        FieldPanel('title'),
+    ]
+
+    @cached_classmethod
```
mateoclarke: I don't really follow whats happening here. Is this all just to enable the Tabs?

joplin/base/models.py
```diff
@@ -70,10 +103,12 @@ def janis_preview_url(self):
 
         return os.environ["JANIS_URL"] + "/en/preview/" + url_page_type + "/" + global_id
 
+    class Meta:
+        abstract = True
 
-class ServicePage(Page, JanisPage):
-    created_at = models.DateTimeField(auto_now_add=True)
-    updated_at = models.DateTimeField(auto_now=True)
+JanisPage._meta.get_field('title').verbose_name='Write an actionable title'
```
mateoclarke: So this line seems like it is composed outside the scope of the class definition. I'm no 🐍 expert but could we do something similar inside the JanisPage class using something like the `self` keyword? 

Also, it seems like an assumption (perhaps a good one) that all our content pages would have actionable titles. But I see that not holding up as we're already talking about Press Releases and Informational page. Might be worth a quick Slack to a content team member to confirm this assumption.

joplin/api/schema.py
```diff
@@ -104,7 +104,7 @@ class ServicePageNode(DjangoObjectType):
     related = graphene.List('api.schema.ServicePageNode')
 
     def resolve_related(self, resolve_info, *args, **kwargs):
-        return self.topic.services.exclude(id=self.id)
+        return self.topic.servicepage_set.exclude(id=self.id)
```
mateoclarke: now that I've seen the Janis PR, why do we need this change?

joplin/base/models.py
ifsimicoded: yea, the problem here is that the title field is initialized in the parent Page object as a field of type CharField -- so I can't just set self.title with a simple string -- it's a specialized CharField class.  I thought about redefining title within the JanisPage model as
```
title = models.CharField(
        verbose_name=_('Write an actionable title'),
        max_length=255,
        help_text=_("The page title as you'd like it to be seen by the public")
)
```
but I thought if the parent definition changed in wagtail in some update, that redefining the title entirely wasn't as update friendly. But I can definitely try this.
But I can reset the verbose name as above, because this line is executed after class definition.

joplin/api/schema.py
ifsimicoded: Hmm--- I'll try to explain this but -- here's a [stackoverflow](https://stackoverflow.com/questions/2642613/what-is-related-name-used-for-in-django) that probably does a better job.

Django allows for reverse lookups of properties on models that are foreign keys to other models. So for example, All JanisPages have a lookup to a topic model, so we can look up all JanisPages that have a particular topic. But since we inherit from JanisPages to create child pagetypes (currently includes ServicePage and ProcessPage) Django needs a way to distinguish between the two types. So give me all ServicePages for a particular topic or give me all ProcessPages for a particular topic. To do this, Django automatically creates a property on a topic model to for this look up by using the <model name>_set. So to access all service pages with a particular topic we would do topic.servicepage_set and similarly for process pages ... topic.processpage_set. So both of those properties are created on the fly by Django. 

The reason for the change in Janis is related to the definition of the topic property in the base JanisPage model. Since we previously didn't inherit the topic property from a base JanisPage model, we could specify the name we wanted for this property (rather than letting Django create it for us with <model name>_set). We did this by specifying a 'related_name' when defining the topic property on each page model as follows ...
```
class ServicePage(Page):
    topic = models.ForeignKey(	
        'base.Topic',	
        on_delete=models.PROTECT,	
        related_name='services',	
        verbose_name='Select a Topic',	
    )
...
```
But since we're now inheriting topic in child page types, I allow Django to autogenerate the topic lookup property name so it will work for all children pages types that inherit from JanisPage.

joplin/base/models.py
ifsimicoded: Oh, don't know how I missed this comment. 
Yep, so lines 82 & 83 is to allow for children models of JanisPage to define the their own TabbedInterface pieces.
lines 85-90 are to create a default tabbed interface for creating/editing content -- where title and topic are included default fields.

ifsimicoded: I'll try to simplify this some more.

joplin/base/models.py
ifsimicoded: Seems like django's not a fan of overriding fields in child classes when the parent isn't an abstract model (Page is not abstract) -- so I'm just going with the suggested approach [here](https://stackoverflow.com/questions/927729/how-to-override-the-verbose-name-of-a-superclass-model-field-in-django)

PR #169, opened 2018-09-19
Address more QA issues

briaguya:
* Make service blocks optional https://github.com/cityofaustin/techstack/issues/692
* Remove delete button from content listing and edit pages https://github.com/cityofaustin/techstack/issues/736
* Add publish functionality to content listing https://github.com/cityofaustin/techstack/issues/693


joplin/base/models.py
```diff
@@ -84,6 +84,7 @@ class ServicePage(Page, JanisPage):
             ('recollect_block', custom_blocks.RecollectBlock()),
         ],
         verbose_name='Add any maps or apps that will help the resident use the service',
+        blank=True
```
briaguya: This makes the map/apps optional

joplin/base/views/joplin_views.py
```diff
@@ -0,0 +1,32 @@
+from django.shortcuts import get_object_or_404, redirect, render
+from wagtail.core.models import Page, UserPagePermissionsProxy
+from wagtail.admin.views import pages
+from wagtail.admin import messages
+from django.utils.translation import ugettext as _
+from django.urls import reverse
+
+def publish(request, page_id):
```
briaguya: This enables publishing from the index page with a view that is very similar to the "confirm unpublish" page

joplin/base/wagtail_hooks.py
```diff
@@ -79,26 +79,33 @@ def joplin_page_listing_more_buttons(page, page_perms, is_parent=False):
...
             priority=20
         )
-    if page_perms.can_delete():
-        yield Button(
-            _('Delete'),
-            reverse('wagtailadmin_pages:delete', args=[page.id]),
-            attrs={'title': _("Delete page '{title}'").format(title=page.get_admin_display_title())},
-            priority=30
-        )
+    # if not page.live:
```
briaguya: This removes the delete option, we don't have any archive functionality yet but once we do we can uncomment this and work towards an archive page flow

joplin/base/views/joplin_views.py
briaguya: I wasn't able to find any functionality in wagtail for publishing based on ID alone, it was based on submitting a page with all of the page data to that endpoint, which doesn't exist on the content list page.

joplin/base/wagtail_hooks.py
briaguya: I agree that it could be slightly problematic, but I think we should address it by making https://github.com/cityofaustin/techstack/issues/694 (implementing archived status) a priority.

joplin/base/wagtail_hooks.py
```diff
@@ -79,26 +79,33 @@ def joplin_page_listing_more_buttons(page, page_perms, is_parent=False):
...
+    #     yield Button(
```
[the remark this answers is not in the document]
briaguya: I'm not sure why we're using the hook except for the fact that wagtail was already using it and now we're overriding it (I think, I may not have a full understanding of it).

As far as `register_page_listing_buttons` vs `register_page_listing_more_buttons` goes, the more buttons are the buttons that go in the "more" dropdown, and are registered via the hook in the more button https://github.com/cityofaustin/joplin/blob/56861d633da9b10ed4c6ce384951f42d370e7bcd/joplin/base/wagtail_hooks.py#L64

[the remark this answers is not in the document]
ifsimicoded: oh, I wasn't clear with my question. I was confused how the built-in hooks relate to our custom hooks since I think both run for the given tags. So how does register_page_listing_buttons relate to our custom hook register_joplin_page_listing_buttons hook -- since I think both hooks are executed in tags that populate the 'wagtailadmin/pages/listing/_buttons.html' partial. And both set of hooks return `return {'page': page, 'buttons': buttons}`. So I'm curious the flow of tags and how order is determined, and how one tag overwrites the other tag.

FWIW, we don't have to use hooks to populate tags to be sent to partials too.

PR #171, opened 2018-09-21
Go straight to edit page from content modal

briaguya:
Instead of using the create page template, we now create the page using an ajax request and then redirect to edit.


joplin/js/CreateContentModal/ButtonBar.js
```diff
@@ -4,15 +4,16 @@ import classNames from 'classnames';
...
   handleBackButton,
   handleNextButton,
   handleCloseButton,
+  onLastStep,
 }) => (
   <div
     className={classNames({
-      ButtonBar: activeStep > 0,
-      'ButtonBar--hidden': activeStep === 0,
+      ButtonBar: !hidden,
```
briaguya: addressing: https://github.com/cityofaustin/joplin/pull/168#discussion_r219266547

joplin/js/CreateContentModal/ChooseTitleStep.js
```diff
@@ -27,11 +28,10 @@ const ChooseTitleStep = ({
     <label htmlFor="page-title" className="ChooseTitleStep__input-label">
       <span className="ChooseTitleStep__input-label--left">Page Title</span>
       <span
-        className={
-          characterCount > maxCharacterCount
-            ? 'ChooseTitleStep__input-label--right ChooseTitleStep__input-label--red'
-            : 'ChooseTitleStep__input-label--right'
-        }
+        className={classNames('ChooseTitleStep__input-label--right', {
```
briaguya: addressing: https://github.com/cityofaustin/joplin/pull/168#discussion_r219259214

joplin/js/CreateContentModal/index.js
```diff
@@ -40,14 +49,14 @@ class CreateContentModal extends Component {
     // Validate title max length
     if (this.state.titleCharacterCount > MAX_TITLE_LENGTH) return false;
 
-    // Skip Topic Select Step for creating a Department
-    if (this.state.type === 'department' && this.state.activeStep === 1) {
-      this.redirectToEditPage();
-      return;
-    }
+    // Validate title min length
+    if (this.state.titleCharacterCount <= 0) return false;
```
ifsimicoded: We probably should alert the user when validation requirements aren't met. Otherwise they may wonder why the Continue button won't work

ifsimicoded: Maybe also validate char count after trimming whitespace?

joplin/js/CreateContentModal/index.js
```diff
@@ -77,13 +85,28 @@ class CreateContentModal extends Component {
     this.setState({ topic: id });
   };
 
-  redirectToEditPage = () => {
-    this.writeToLocalStorage();
-    window.location.href = this.state.redirectUrl;
+  redirectToEditPage = id => {
+    window.location.href = `/admin/pages/${id}/edit/`;
```
ifsimicoded: since we don't really reuse this redirect method, could we just kill the method and put the redirect line in the then block on line 104?

joplin/js/CreateContentModal/index.js
```diff
@@ -77,13 +85,28 @@ class CreateContentModal extends Component {
...
+          title: this.state.title,
+          topic: this.state.topic,
+        },
+        { headers: { 'X-CSRFToken': Cookies.get('csrftoken') } },
+      )
+      .then(response => {
+        this.redirectToEditPage(response.data.id);
+      })
+      .catch(error => {
+        console.log(error);
```
ifsimicoded: Same here with error states. We probably should alert the user when the create page fails and provide them an alternative?

joplin/js/CreateContentModal/index.js
```diff
@@ -77,13 +85,28 @@ class CreateContentModal extends Component {
...
+      })
+      .bind(this);
```
ifsimicoded: 💅 didn't know about bind!

joplin/js/CreateContentModal/ChooseTypeStep.js
```diff
@@ -2,53 +2,53 @@ import React from 'react';
...
+            handleTypeSelect({
+              type: 'process',
+            })
+          }
         >
-          <img
-            src={processPageImage}
-            alt="Service Page"
-          />
+          <img src={processPageImage} alt="Service Page" />
```
ifsimicoded: alt text should be Process Page

joplin/urls.py
```diff
@@ -23,6 +23,7 @@ def home(request):
 
     path('admin/', home),
     url(r'admin/pages/(\d+)/publish/$', joplin_views.publish, name='publish'),
+    url(r'admin/pages/new_from_modal/$', joplin_views.new_page_from_modal, name='new_page_from_modal'),
```
ifsimicoded: nitpick, maybe we could create a path for admin api routes. so like admin/joplinapi/pages/create or admin/ajax/pages/create -- or something similar?

PR #172, opened 2018-09-25
Get Travis stuff into prod

briaguya:


.travis.yml
```diff
@@ -0,0 +1,83 @@
...
+# We will need these services:
+services:
+- docker
+- postgresql
+
+sudo: required
+
+# We need PostgreSQL 10 installed
+addons:
+  postgresql: "10"
```
ifsimicoded: is this ok, given the change to 9.6 in your other branch?

.travis.yml
```diff
@@ -0,0 +1,83 @@
...
+  apt:
+    packages:
+      - postgresql-10
```
ifsimicoded: same here -- postgres 9.6?

.travis.yml
sergiogcx: I have fixed this in my branch, we will continue to use 10. A specific step was added in the dockerfile to install Pg10.

sergiogcx: I working to get this sorted before Thursday

PR #174, opened 2018-09-27
Bs fancysteps dbstuff

briaguya:


docker-compose.local.yml
```diff
@@ -1,7 +1,7 @@
 version: '3'
 services:
   joplindb:
-    image: postgres
+    image: postgres:9.6
```
briaguya: dbbackup had issues with version mismatches using postgres 10 in this container, using 9 fixed it. I didn't spend too much time investigating how to get it so we had the postgres 10 client running in our joplin container with wagtail, if we find we need postgres 10 in the future that should be doable

docker-entrypoint.sh
```diff
@@ -24,4 +24,9 @@ fi
 # Run Data Migrations Only (Static files can remain local)
 python ./joplin/manage.py migrate --noinput
 
+if [ "$LOAD_DATA" == "on" ]; then
```
briaguya: Instead of loading data from fixtures files, or using a broken migration script with hardcoded data, we can simply load the latest backup (assuming it was created using the current migrations)

joplin/base/migrations/0031_auto_20180925_2309.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 2.0.8 on 2018-09-25 23:09
```
briaguya: not sure how this hadn't made it into master yet

joplin/base/migrations/0032_servicepage_steps.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 2.0.8 on 2018-09-27 04:54
```
briaguya: Here we begin the migration process, our first step is to add the new field without deleting the old steps relation. This lets us do a data migration from one to the other.

joplin/base/migrations/0033_auto_20180927_0500.py
```diff
@@ -0,0 +1,37 @@
+# Generated by Django 2.0.8 on 2018-09-27 05:00
```
briaguya: This is the data migration script, I created it using `python manage.py migrate base --empty` to initialize it without anything in it.

joplin/base/migrations/0033_auto_20180927_0500.py
```diff
@@ -0,0 +1,37 @@
...
+
+from django.db import migrations
+from pprint import pprint
+from wagtail.core.blocks.stream_block import StreamValue
+
+def better_steps(apps, schema_editor):
+    '''
+    We can't import the models directly as they may be a newer
+    version than this migration expects. We use the historical versions.
+    '''
```
briaguya: This comment is decently self explanatory, but to elaborate - this allows us to avoid importing models (which change) and instead use the model that is currently in the DB. In this specific scenario:

Original model - Orderable steps relation only
Final model - StreamFieldSteps only
Model at this point in time in the DB schema migration process - 🌟BOTH 🌟

joplin/base/migrations/0033_auto_20180927_0500.py
```diff
@@ -0,0 +1,37 @@
...
+    ServicePage = apps.get_model('base', 'ServicePage')
+    ServicePageStep = apps.get_model('base', 'ServicePageStep')
+
+    for page in ServicePage.objects.all():
+      pageSteps = []
+      for step in ServicePageStep.objects.all():
+        if step.page_id == page.id:
```
briaguya: I know there's a cleaner way to do this than iterating through everything and using an if statement, I was just tired of looking through the django docs by this point and it was working

joplin/base/migrations/0033_auto_20180927_0500.py
```diff
@@ -0,0 +1,37 @@
...
+          streamStep = {
+            'type': 'basic_step',
+            'value': step.step_description
+          }
```
briaguya: For this migration I'm just taking every step we had in the old model and moving it into the new model as a basic step. We're losing our translations moving to a streamfield here so I'm just using the standard description instead of each of the translated ones.

joplin/base/migrations/0033_auto_20180927_0500.py
```diff
@@ -0,0 +1,37 @@
...
+          pageSteps.append(streamStep)
+      stream_block = page.steps.stream_block
+      page.steps = StreamValue(stream_block, pageSteps, is_lazy=True)
```
briaguya: This part is kind of weird, I followed the pattern I found [here](https://github.com/wagtail/wagtail/issues/2110#issuecomment-187821699), I'll do my best to explain it:

A streamfield has a StreamValue, which needs:
* stream_block - Still don't fully understand what it is, but since we have already initialized the StreamField on the ServicePage, we can just copy it back when we rebuild the steps.
* stream_data - In our case this is an array of steps, hence using pageSteps here.
* is_lazy - Again, not completely sure what this does, but it was already set to True in all of our steps StreamFields , so I'm making sure it stays that way.

joplin/base/migrations/0034_auto_20180927_0731.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 2.0.8 on 2018-09-27 07:31
```
briaguya: Now that we've copied all the steps from the old model to the new one, we can go ahead and remove the old field/model

joplin/base/models.py
```diff
@@ -111,6 +110,35 @@ class Meta:
 class ServicePage(JanisPage):
     janis_url_page_type = "services"
 
+    steps = StreamField(
```
briaguya: Thanks for getting this working @ifsimicoded !

joplin/db/backups/default-58807f2ec735-2018-09-27-073204.psql
```diff
@@ -0,0 +1,5612 @@
+--
```
briaguya: This is the only DB backup we actually need, as it's the most recent.

1. The first backup was made from master before any changes
2. Then I made one after adding the new steps model without removing the old one and running the schema migrations for that.
3. Then I made another one after creating and running the data migrations to get the steps from the old model to the new one.
4. That's this one! It's from the DB once we removed the old steps model.

scripts/serve-local.sh
```diff
@@ -2,22 +2,6 @@
 
 set -o errexit
 
-DB_FILE='./joplin/db.sqlite3'
```
briaguya: Just a little basic cleanup here

sergiogcx:
The django migration depends on the images or it will fail, everything else can be deleted.

Dockerfile
```diff
@@ -16,7 +16,6 @@ EXPOSE $PORT
 RUN mkdir /app
 WORKDIR /app
 
-COPY "$PWD/fixtures" /app/fixtures
```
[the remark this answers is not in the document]
briaguya: put these back because images

briaguya:
@ifsimicoded basic steps for getting the translation parts working:
* First delete a couple backups: I think you only want this one: default-51d5350ff249-2018-09-27-065431.psql (if you go back through the commit history we should be able to confirm this one got made at the same time as migration 0032_servicepage_steps.py
* Delete all migrations after 0032_servicepage_steps.py
* Start up joplin with LOAD_DATA=on
* ***WHILE IT'S RUNNING!***
  * Update the ServicePage model to have all the translation versions of the steps
  * Make a new migration
  * ***Run MIGRATE***
  * Make a new backup

joplin/base/migrations/0033_auto_20180927_0500.py
ifsimicoded: Can you add this info to the readme -- basically how to create a migration process.

PR #178, opened 2018-10-04
Postgres All the Things

sergiogcx:
This allows us to create PRs with PostgreSQL and migration processes using Travis.


.travis.yml
```diff
@@ -2,11 +2,24 @@ language: python
...
+#       - git commit message contains "[joplin:build]"
+#
+#   Note: PR_AUTOBUILD or PR_AUTODEPLOY variables can be set either in Travis > Settings or
+#   at the env: global: section of this configuration file. See example below.
+#
+
+if: |
+  (branch IN (master, production) AND type = push) OR \
+  (type = pull_request AND env(PR_AUTOBUILD) IN (true, enabled, on)) OR \
+  (type = pull_request AND env(PR_AUTODEPLOY) =~ concat(head_branch) ) OR \
```
briaguya: Do we set PR_AUTOBUILD and PR_AUTODEPLOY to true by default?

.travis.yml
```diff
@@ -2,11 +2,24 @@ language: python
...
+  (type = pull_request AND commit_message =~ /\[joplin\:build\]/ )
```
briaguya: Do we want this override? I think it makes more sense to remove it.

.travis.yml
```diff
@@ -17,22 +30,25 @@ sudo: required
...
 # Environment Vars
 env:
   global:
+  - PR_AUTOBUILD: "true" # (example, enables automatic PR build when present)
   - PATH=$HOME/.local/bin:$PATH
-  - AWS_DEFAULT_REGION: us-east-1
-  - AWS_BUCKET_BACKUPS: joplin-austin-gov-archive
-  - JOPLIN_TARGET: joplin-personal
   - JOPLIN_IMAGE_NAME: joplin-base
+  # HEROKU_API_KEY="..."
```
briaguya: Unfortunate thing here, if we add another secure key, then these comments will immediately disappear. Not a real problem, just a good thing to note.

Dockerfile.local
```diff
@@ -2,14 +2,6 @@ FROM joplin-base
 
 ENV DEPLOYMENT_MODE "LOCAL"
 
-
-RUN apt-get update \
```
briaguya: Why don't we get the updated postgres here anymore? I'm sure I'm just missing something.

sergiogcx: It's already installed in Dockerfile.base, which is used here at "FROM joplin-base"

.travis.yml
sergiogcx: Maybe these should be documented in the Readme file

.travis.yml
sergiogcx: Do you mean the "[joplin:build]" keyword? I personally like the convenience it gives me to build without having to change code or changing environment variables, especially if I only need one quick build. 

It would be very difficult to type that by accident, and it would only work for PRs.

.travis.yml
briaguya: It seems like this is changing the logic away from what we had before.

It used to be:
* Work in a branch, commit as much as you want, never deploys
* Make a PR, automatically have a PR deployment
* Make changes in the branch you have a PR for, automatically updates the PR environment

Could you explain the reasoning behind adding this complexity?

.travis.yml
sergiogcx: Got it, I just set up the new logic:
(branch IN (master, production) AND type = push) OR (type = pull_request)

Meaning it will build all updates to PRs, master and production branches, leaving regular branches alone.

.travis.yml
sergiogcx: Got it, I just set up the new logic:
(branch IN (master, production) AND type = push) OR (type = pull_request)

Meaning it will build all updates to PRs, master and production branches, leaving regular branches alone.

PR #183, opened 2018-10-23
Merge to Production

sergiogcx:
These are the changes made to master (staging) and proposed to be made to production.


sergiogcx:
The working PR is https://joplin-staging-pr-183.herokuapp.com

amenity:
@toribr should also review

sergiogcx:
I will be creating a PR directly from master next time (as opposed to sg-joplin-devops-prapps-prod). I decided to do it this way because I cannot predict with exactitude how the deployment will behave given that this is both master and a PR.

From what I see in the code, a new review app will be created using the same format (joplin-staging-pr-183.herokuapp.com). I am looking to make sure all functions in the script identify master PRs as PRs and not as master, that way we do not use branches like the one I created.

PR #184, opened 2018-10-24
Service Page UI Fixes

briaguya:
Everything in https://github.com/cityofaustin/techstack/issues/834 not covered by https://github.com/cityofaustin/techstack/issues/836 or https://github.com/cityofaustin/techstack/issues/837

### DIFF
--- a/joplin/css/editor.scss
+++ b/joplin/css/editor.scss
@@ -55,3 +55,7 @@ footer {
 fieldset {
   max-width: none !important;
 }
+
+label {
+  width: auto !important;
+}
--- a/joplin/css/editor.scss
+++ b/joplin/css/editor.scss
@@ -32,7 +32,7 @@ footer {
 .icon-help-inverse.show {
   display: inline;
   float: right;
-  margin: .5rem 4rem 0 0;
+  margin: 0.5rem 4rem 0 0;
 }
 
 .actions {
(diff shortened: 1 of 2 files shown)


joplin/css/editor.scss
```diff
@@ -51,3 +51,11 @@ footer {
 .editor__save-button {
   min-width: 130px;
 }
+
```
ifsimicoded: hey @briaguya a couple things
1) I'm not sure what this is doing -- I comment them out and the edit page input titles and fields look the same. You'll see in admin.scss (our customizations) we have 
```
.object fieldset,
.object .field-content {
  width: 100%;
}
```
and page_editor.css (part of wagtail css) there is
```
.object>h2 label {
    display: inline;
    text-transform: inherit;
    font-weight: inherit;
    float: none;
    width: auto;
    color: inherit;
    font-size: inherit;
}
```
2) using !important worries me in that we're making a nightmare for those adjusting styles later. Can we better systematize and scope our style overrides some how so we don't end up writing scss over even our own scss?

briaguya: @ifsimicoded if we're on a large screen we end up with layout issues:
Staging:
screen shot 2018-10-30 at 12 40 34 pm
This branch:
screen shot 2018-10-30 at 12 40 59 pm

I'm not sure how/where to better address this (but I agree that !important is less than ideal)

PR #186, opened 2018-10-25
Reckless hotfixes

briaguya:


joplin/api/schema.py
```diff
@@ -108,7 +108,7 @@ def resolve_related(self, resolve_info, *args, **kwargs):
 
     class Meta:
         model = ServicePage
-        filter_fields = ['id', 'slug', 'topic', 'topic__slug']
+        filter_fields = ['id', 'slug', 'topic', 'topic__slug', 'live']
```
ifsimicoded: Hey if we limit the request to just return live pages all the time we won't have to make updates to the Janis repo. we could do this by adding to the Query class in this file.
```
def resolve_all_service_pages(self, info):
        return ServicePage.objects.live()

def resolve_all_processes(self, info):
        return ProcessPage.objects.live()
...
```
etc for all models?

joplin/settings.py
```diff
@@ -271,6 +271,8 @@
     DEFAULT_FILE_STORAGE = 'custom_storages.MediaStorage'
     DBBACKUP_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
 
+JANIS_SLUG_URL=""
```
ifsimicoded: What's this for? I know we use this url to trigger rebuild of janis when a page is unpublished or published on staging and prod, if all the settings variables are defined in signals.py (below).
```
if not all([settings.HEROKU_KEY, settings.HEROKU_JANIS_APP_NAME]):
```
should we just add `setting.JANIS_SLUG_URL` to the array above instead of making an empty definition?

PR #187, opened 2018-10-26
Prevent creating duplicate pages with modal

briaguya:
This simply hides the modal while we're waiting to redirect to the newly created page edit screen. We still have the rest of the page greyed out during this time.

### DIFF
--- a/joplin/js/CreateContentModal/index.js
+++ b/joplin/js/CreateContentModal/index.js
@@ -170,15 +170,13 @@ class CreateContentModal extends Component {
                         themeTopicTree={THEME_TOPIC_TREE}
                       />
                     )}
-                    {!this.state.creatingContent && (
-                      <ButtonBar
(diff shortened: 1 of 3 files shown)


briaguya:
@ifsimicoded Good catch on the spinner. I had forgotten that we already had an issue for that and was going at this in "bug fix" mode instead of looking for a well designed solution. I've talked to @amenity and we're moving forward with the fix including finishing up https://github.com/cityofaustin/techstack/issues/743

Also, would you like to take a stab and breaking up the bootstrap/react combo weirdness or should I?

ifsimicoded:
Nice! If you’ve got bandwidth, do it!!!

briaguya:
@ifsimicoded I'll hop on it!

briaguya:
Hey @ifsimicoded I've addressed everything you mentioned:
gif of it

PR #225, opened 2019-05-03
1842 translate service page steps

briaguya:


easherma:
spanish
Appears to be working on janis + joplin local.

PR #236, opened 2019-05-17
Merge master into prod

sergiogcx:
We need to merge master into prod, this is to be able to build janis on-demand. The mechanisms are there to specify which API the builder task needs to run against, the CloudFront distro to invalidate cache, and the bucket where to deploy the built files.

If you wish to play with this builder, try publishing or un-publishing an article on joplin-staging.

http://joplin-staging.herokuapp.com

Once published, you can see the progress in slack, in the channel `#janis-builds`

And the output in this bucket:

http://janis-lab-staging.s3-website-us-east-1.amazonaws.com/


sergiogcx:
Actually, there is a small fix I need to work on in master. Closing for now, will try again on Monday.

PR #245, opened 2019-05-23
1944 account component

easherma:
Fixes https://github.com/cityofaustin/techstack/issues/1944

matches design spec with some tweaks

the css is pretty messy, there are some dangled relationships between all the stuff joplin is pulling in
this leverages bootstrap 3 dropdown method since that is already dependency


joplin/css/layouts/headers.scss
```diff
@@ -1,43 +1,43 @@
...
-  padding: 0.75rem 1.5rem;
-  background: $coa-color-darkest-gray;
-  position: fixed;
-  z-index: 3;
-  width: 100%;
+    display: flex;
+    padding: 0.75rem 1.5rem;
+    background: $coa-color-darkest-gray;
+    position: fixed;
+    z-index: 4;
```
easherma: puts it on the same level as the topmost bar

PR #250, opened 2019-05-27
1930 share draft ui

easherma:
fixes https://github.com/cityofaustin/techstack/issues/1930


easherma:
ehh this got messy with some commits, see https://github.com/cityofaustin/joplin/pull/251

PR #260, opened 2019-05-30
2193 up page description limit

easherma:
Fixes https://github.com/cityofaustin/techstack/issues/2193

I did:

* update the data model
* made migrations
* ran migrations

I tried running test migrations but ran into an error trying to pull the image because I wasn't logged in to docker. Not clear if this last step is 100% required or not?


easherma:
Nick says:

"If you don’t update the backup file after running a migration, then future LOAD_DATA=on commands could fail. You should finish the whole migration-test process every time you change migrations"

PR #262, opened 2019-05-31
1376 login case insensitive

easherma:
Fixes https://github.com/cityofaustin/techstack/issues/1376

@briaguya we may want to test this more eventually 

I'm using this: 
https://docs.djangoproject.com/en/2.2/ref/models/querysets/#iexact

This should work for now. But I'm not sure how we'd handle sign-up/registration of emails. We have a unique validator there, but I'm not sure if that validator is also case sensitive? Another route to go might be casting all emails to lower() when making accounts for people, etc.

### DIFF
--- a/joplin/users/models.py
+++ b/joplin/users/models.py
@@ -10,6 +10,13 @@ class UserManager(BaseUserManager):
 
     use_in_migrations = True
 
+    def get_by_natural_key(self, username):
+        return self.get(username__iexact = username)
+
+    # def get_by_natural_key(self, username):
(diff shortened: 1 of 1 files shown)


joplin/users/models.py
```diff
@@ -10,6 +10,9 @@ class UserManager(BaseUserManager):
 
     use_in_migrations = True
 
+    def get_by_natural_key(self, email):
```
Niicck: Add Documentation describing what this solves

PR #264, opened 2019-05-31
2230 fix search

briaguya:
Fixes https://github.com/cityofaustin/techstack/issues/2230

### DIFF
--- a/joplin/base/wagtail_hooks.py
+++ b/joplin/base/wagtail_hooks.py
@@ -69,7 +69,7 @@ def joplin_page_listing_buttons(page, page_perms, is_parent=False):
             attrs={'title': _("Preview draft version of '{title}'").format(title=page.get_admin_display_title()), 'target': '_blank'},
             priority=20
         )
-    if page.live and page.url:
+    if page.live and page.url and hasattr(page, 'janis_url'):
         yield PageListingButton(
             _('View live'),
             page.janis_url(),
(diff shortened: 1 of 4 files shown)


joplin/base/wagtail_hooks.py
```diff
@@ -69,7 +69,7 @@ def joplin_page_listing_buttons(page, page_perms, is_parent=False):
             attrs={'title': _("Preview draft version of '{title}'").format(title=page.get_admin_display_title()), 'target': '_blank'},
             priority=20
         )
-    if page.live and page.url:
+    if page.live and page.url and hasattr(page, 'janis_url'):
```
easherma: Could we leave a comment here about what `janis_url` is and what its being used for in this context?

joplin/templates/wagtailadmin/pages/listing/_table_headers_explore.html
```diff
@@ -29,7 +29,7 @@
     {% endif %}
     <th class="title">
         {% if sortable %}
-            <a href="{% url 'wagtailadmin_explore' parent_page.id %}?ordering={% if ordering == 'title' %}-{% endif %}title" class="icon icon-arrow-{% if ordering == 'title' %}down-after{% elif ordering == '-title' %}up-after{% else %}down-after{% endif %} {% if ordering == 'title' or ordering == '-title' %}teal{% endif %}">
+            <a href="{% url 'wagtailadmin_explore' 3 %}?ordering={% if ordering == 'title' %}-{% endif %}title" class="icon icon-arrow-{% if ordering == 'title' %}down-after{% elif ordering == '-title' %}up-after{% else %}down-after{% endif %} {% if ordering == 'title' or ordering == '-title' %}teal{% endif %}">
```
easherma: Might be worth leaving a note here/linking to where we talk about hard-coding?

Niicck: Is there a way to do this without hardcoding the page_id?

briaguya: the issue I found with using a non-hardcoded id is that we don't always have a parent_page in here, but we still want to have the explore button work.

We currently have every page as a child of "Home". We could have some logic to say "if we have a parent page, use that ID, if not default to 3" but if we end up with a parent_page other than "Home" a bunch of other stuff would break anyways.

PR #267, opened 2019-06-04
Issue 1959: translate previews and sharing

Niicck:
Surprisingly big issue, I'd love to have ya'll thoughts on this. If we like this pattern, we can use it going forward.

Ref: https://github.com/cityofaustin/techstack/issues/1959
To QA: run it locally. See if you can do a "Mobile Preview", "Preview", and "Share" in Spanish.

Bonus changes:
- gunicorn reloads, so you can see python changes without restarting your container
- node-sass re-compiling happens automatically in the assets docker container
- the big one: we now have any easy way to move django objects into javascript

From edit.html:
```
<script>
      // Load this before render_bundle to use django vars inside javascript
      var previewUrlData = {};
      try {
        // get python dictionary into valid JSON format
        previewUrlData=JSON.parse("{{page.preview_url_data|escapejs}}".replace(/\'/g, '\"'));
      } catch (e) {}
      var djangoData = {
        previewUrlData: previewUrlData,
        fallBackPreviewUrl: "{{ page.fallback_preview_url }}",
        styleGuideUrl: "{{ style_guide_url }}"
      }
    </script>
    {% render_bundle 'editor' %}
```

There are a couple benefits to this approach rather than saving values in hidden <div> tags. One, explicit javascript variables become the source of truth for values, which makes variable management easier to follow and manipulate. Two, we can pass entire JSON objects from django models/views besides strings. Three, variable declaration can be done in one place in the django template, which makes it easier to find stuff and know what you can work with.

We also now have a "state" object in editor.js to group together dynamically changes variables. Naturally, it's not as robust as react and can't do automatic component re-renders or whatever. But it's something! This issue has been fun. I kind of understand better why react was made.

I think these changes could go a long way to making django behave more like a web app. Which we'll need more of as our feature requests start asking for more dynamic stuff.

### DIFF
--- a/app.Dockerfile
+++ b/app.Dockerfile
@@ -10,11 +10,6 @@ RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ stretch-pgdg main' >  /et
     && curl -s https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - \
     && apt-get update && apt-get install -y postgresql-client
 
-# Install nodejs dependencies
-RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
-RUN apt-get update; apt-get -y install nodejs
-RUN npm install --global yarn
-
 # Set Environment Variables
 ENV PYTHONUNBUFFERED=1
 ENV WEB_CONCURRENCY=4
@@ -50,6 +45,11 @@ CMD ["gunicorn", "joplin.wsgi:application", "--pythonpath", "/app/joplin", "--re
 # joplin-base => joplin-deployed
 
 FROM joplin-base as joplin-deployed
+
+# Install nodejs dependencies
+RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
+RUN apt-get update; apt-get -y install nodejs
+RUN npm install --global yarn
 
 # Build nodejs dependencies for deployed builds
 WORKDIR /app/joplin
--- a/app.Dockerfile
+++ b/app.Dockerfile
@@ -35,7 +35,7 @@ ENV DEPLOYMENT_MODE "LOCAL"
 # Run Migrations
 ENTRYPOINT ["./docker-entrypoint.sh"]
 # Start the Joplin server
-CMD ["gunicorn", "joplin.wsgi:application", "--pythonpath", "/app/joplin"]
+CMD ["gunicorn", "joplin.wsgi:application", "--pythonpath", "/app/joplin", "--reload"]
 
 ########################################################
 # joplin-base => joplin-deployed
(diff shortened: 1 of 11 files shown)


joplin/templates/wagtailadmin/shared/sidebar-preview.html
```diff
@@ -1,19 +1,17 @@
...
-                    href="{{ page.janis_preview_url }}" 
-                {% else %}
-                    {# TODO: When we are creating a page (as opposed to editing), we do not yet have a saved revision so what do we show in the mobile preview? #}
-                    href="https://alpha.austin.gov" 
-                {% endif %}
-                target="sidebar-iframe"
+            <a id="mobile-preview-sidebar-button" class="coa-sidebar-button js-toggle-active-panel"
+              {# Placeholder that will be overwritten by changeLanguage() in editor.js #}
+              {# When editing an existing page, there will be a mobile preview #}
+              {# However, there will not be a mobile preview yet for a new page #}
```
briaguya: I don't think we actually have "new" pages in the sense wagtail thinks of them. Our create content modal adds some basic data to a page and then saves it before redirecting to a "edit" page screen

app.Dockerfile
```diff
@@ -10,17 +10,26 @@ RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ stretch-pgdg main' >  /et
     && curl -s https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - \
     && apt-get update && apt-get install -y postgresql-client
 
-COPY deploy/requirements.txt /deploy/requirements.txt
-RUN pip install --no-cache-dir --disable-pip-version-check --requirement /deploy/requirements.txt
+# Install nodejs dependencies
+RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
```
briaguya: This is great for now, but it'd be cool to think about how we want this to work in a deployed environment. Might be worth trying to make a more lightweight version where we don't need node running so we can just use pre-built webpack bundles to handle this stuff. (definitely a separate issue)

Niicck: Worth pondering how to make is more lightweight. We could build + compile node on the CircleCI containers rather than on the joplin container itself. 

Our current deployed environments already do install node. The change I made to the dockerfile just installs node on our local containers as well. 

That was actually a misfired attempt to do something else. I thought that sass had to be rebuilt on the joplin_app container rather than the joplin_asset container.

joplin/templates/wagtailadmin/shared/sidebar-preview.html
Niicck: Huh, okay. I'll take out my misleading comments.

PR #268, opened 2019-06-04
2212 helper text location

easherma:
Fixes https://github.com/cityofaustin/techstack/issues/2212

This could potentially be a big change, so I did it in as incremental a way as possible. 

BrainDump:

- The way we use `verbose_name` is clever but not at all standard for the way [Django expects it](https://docs.djangoproject.com/en/2.2/ref/models/options/#verbose-name). TL;DR Django expects this to refer to the name of the object, not a... sentence describing what the object is for. Maybe we could/should use docstrings or some other self-documenting thing? Eventually we should change it (IMO), but I was worried about that having ripple effects elsewhere in the codebase.
- This branch makes use of the [Wagtail Help Panel](http://docs.wagtail.io/en/v2.5.1/reference/pages/panels.html#helppanel), it's pretty groovy
- Because of the above concerns, I called verbose_name and help_text props in the multifield panel heading and help panel respectively. This led to duplicate copy in the edit interface, so I hid those labels for now.


joplin/base/migrations/0073_auto_20190604_2124.py
```diff
@@ -0,0 +1,39 @@
+# Generated by Django 2.2.2 on 2019-06-04 21:24
```
briaguya: There are a lot of migrations in here, any chance you could do a little bit of git time travel and:
1. Run master locally
2. Pull your branch, delete all migrations
3. Make/run migrations

If that ends up being a mess don't worry about it, just figured it might be a nice way to make these more parseable.

joplin/base/models.py
```diff
@@ -149,6 +149,7 @@ class ServicePage(JanisPage):
             )),
         ],
         verbose_name='Write out the steps a resident needs to take to use the service',
+        # this gets called in the help panel
```
briaguya: 👍

joplin/base/models.py
```diff
@@ -247,6 +263,7 @@ class InformationPage(JanisPage):
         features=WYSIWYG_GENERAL,
         verbose_name='Write any additional content describing the service',
         blank=True
+
```
briaguya: nitpick line

joplin/css/admin.scss
```diff
@@ -22,8 +22,8 @@
 
 .translation-tab .object .object-help {
     width: auto;
-    opacity: 1;
-    position: static;
+    // opacity: 1;
```
briaguya: do we plan on using these again in the future for something? if not let's just delete them instead of commenting them out

joplin/css/editor.scss
```diff
@@ -9,25 +11,21 @@
...
 }
 
-.object-help.help {
-    opacity: 1;
-    width: auto;
-    float: left;
-    position: relative;
-    padding: 0;
-}
+// .object-help.help {
```
briaguya: let's delete!

joplin/css/pages/edit.scss
```diff
@@ -30,6 +30,26 @@
...
 
+.coa-multiField-nopadding li {
+    padding: 0;
+}
+
+.coa-helpPanel fieldset {
+    padding: 0;
+}
+
+// tech debt: hiding help instead of removing, check scoping, bad idea?
```
briaguya: I'm not too worried about it, this seems like a good enough solution (unless we start running into weird visual bugs I haven't seen yet

PR #272, opened 2019-06-05
add make migrations step that merges conflicts

easherma:
So the idea here:

* docker entrypoint is used every time we spin up a container
* so every time we run migrations....
* so _before_ that every time lets makemigrations, with the --merge flag to deal with conflicts

This should help catch missing migrations or conflicts when we merge to master. 

Cons:
* I'm not as familar of the overall build process, so not sure if this has unintended consequences?
* Migrations will look kinda messy, but I think this is more compatible than deleting/rolling back
  * With this in mind, we could consider squashing all the migrations every time we ship a new major/minor version?
  * I played around with adding the branch + commit hash to the name of the migration for reference, but I think this is overkill for now

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -20,10 +20,6 @@ if [ $DEPLOYMENT_MODE == "LOCAL" ]; then
   >&2 echo "Postgres is up - executing command"
 fi
 
-# makes migrations, with merge flag to handle conflics
-echo "making Schema Migrations"
-python ./joplin/manage.py makemigrations --merge --noinput
-
 # Run schema Migrations Only (Static files can remain local)
 echo "Running schema Migrations"
 python ./joplin/manage.py migrate --noinput
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -20,8 +20,12 @@ if [ $DEPLOYMENT_MODE == "LOCAL" ]; then
   >&2 echo "Postgres is up - executing command"
 fi
 
-# Run Data Migrations Only (Static files can remain local)
-echo "Running Data Migrations"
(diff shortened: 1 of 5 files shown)


docker-entrypoint.sh
```diff
@@ -20,8 +20,12 @@ if [ $DEPLOYMENT_MODE == "LOCAL" ]; then
   >&2 echo "Postgres is up - executing command"
 fi
 
-# Run Data Migrations Only (Static files can remain local)
-echo "Running Data Migrations"
+# makes migrations, with merge flag to handle conflics
+echo "making Schema Migrations"
+python ./joplin/manage.py makemigrations --merge --noinput
```
briaguya: I'm worried that this is a little too aggressive. One main concern is if we made some changes while running locally and everything seems fine, but then forgot to shut down and restart the container. Then the migration might get made in a deployed environment and not checked in to code. Later on, it'd be possible to have our migrations out of sync between deployed environments and our codebase.

It's possible I'm overthinking this, but right now this seems scary.

easherma: I get that line of thinking, it might be best if we could try to actually create that scenario and see what happens. 

Another thing I was thinking is that we could run 
`python ./joplin/manage.py makemigrations --check`
Which checks if there are model changes that have no migrations for them. We could put that in CI testing or also in this entrypoint?

The reason why I left it out was mainly thinking 
'What would we want to do if there weren't migrations made for a model change?'
and I couldn't think of cases where we wouldn't then want to make a migration, even if we ended up later committing a new migration that undid that change.

easherma: @Niicck and I made this less scary!

PR #276, opened 2019-06-11
example POC for forms as a model

easherma:
hack to make a form atm: 
goto http://localhost:8000/admin/pages/3/add_subpage/

Currently this breaks:
1) When you go to save a draft of a form
2) Once you have this draft, viewing admin gets broken mainly because FormPage types don't have a janis_preview_url

1) We don't really have docs covering what cases to check for/how to add NEW models to joplin
2) We should be able to put in some error handling to handle these cases more gracefully


briaguya:
Screen Shot 2019-06-12 at 10 39 43 AM
@easherma assuming this was the error you were talking about, just committed to the branch with a quick fix.

easherma:
This branch was used successfully for our bake-off, so taking off the WIP label but assuming we won't merge it for now.

PR #278, opened 2019-06-20
[wip] re-write routes for styleguide pages

easherma:


easherma:
btw, another option for implementing this could be using help panels to have the content be directly on joplin.

PR #282, opened 2019-06-25
2311 better debug

easherma:
Using DEBUG_TOOLBAR=on will add the django debug toolbar
What remains is figuring out a good way to set the value in CircleCI or with Heroku
For now you could set it for your PR app in Heroku and rebuild and the toolbar should show up. (haven't tested that yet)


easherma:
I haven't had time to make the changes @Niicck suggested

easherma:
@Niicck I'll make a deal with you; if I open up a branch splitting up settings.py by dev environment if you'll pair with me to set up the flags/changes you mentioned above. I'm not super experienced writing bash scripts.

PR #293, opened 2019-07-05
[wip] 2271 related departments

easherma:
Fixes https://github.com/cityofaustin/techstack/issues/2271 , implementation for Joplin 

Todo:
Check how this influences the Graphql schema + Janis builds


easherma:
Sounds good!

TODO:

- [ ] test endpoint for janis with the code as is
- [ ] refactor graphene stuff as needed, maybe make into it's own issue if needed

easherma:
So making this show up in GraphQL has been a challenge, I think mainly because of the way I'm adding it to generic Page instead of a specific class we created? Not 100% sure.

Overall I think our overall model layout could be improved, and our Graphene implementation is brittle. Both are worthy challenges to tackle but might be out of scope for MVP.

easherma:
See notes. Closing for now. This might be a nice thing to revisit later to make our code more elegant

PR #299, opened 2019-07-17
2510 kitchen sink

easherma:
this is mainly for discussion purposes, to show a couple options we have for nesting items on the lefthand menu

1) We resurface the default snippets menu, which lists all the snippets
2) We make a custom submenu of some of the important items we've identified. We could group this into this folder instead of listing them all on the side.
3) We resurface the default settings menu to show a sense of what stuff can be modified by default that content or UX might find useful. The construction of this menu can be somewhat customized and will be further customizable in Wagtail 2.6

All of these buttons and menus can be customized to only show for certain types of users.


briaguya:
Screen Shot 2019-07-17 at 6 17 47 PM
silly menu

easherma:
yes it is sorta intentionally silly

PR #300, opened 2019-07-18
Master --> Prod (Joplin 2.1.1 I'd call it)

easherma:
* adds timeout
* routes styleguide to correct content type (for those that have one)
* our custom publish view always saves a revision to hopefully avoid some data bugs (still wip on a better fix)
* left sidebar menu has been refactored
* material icons have been added
* django extensions added for some nifty shell/model stuff 
* djano autodocumentation as been added to a route for referencing models, etc


briaguya:
Just ran prod janis against staging joplin, everything seems fine so I'm going to do a prod push (on a Friday 🎃👻) and make sure everything is still in good shape

PR #301, opened 2019-07-22
Issue 2311: fix revisions in seeding data

Niicck:
Ref: https://github.com/cityofaustin/techstack/issues/2311

Prod revisions were incompatible with our local/dev data. Now they are. Now you can publish anything anywhere. Hooray! 

Check out the patched sql job in `scripts/sanitize_revision_data.sql` to see how it works.

Deployed at https://joplin-pr-2311-bad-data.herokuapp.com/admin


scripts/serve-local.sh
```diff
@@ -86,5 +86,6 @@ else
   # Only run containers for db and assets
   docker-compose -f docker-compose.yml -f docker-compose.local_override.yml up -d db assets
 
+  sh $CURRENT_DIR/../scripts/setup-undockered-data.sh
```
easherma: Could we have a comment here about why this script is called here? Looks like the other script is running existing migrations locally and then also loading data, irregardless of the other flags you set up fro serve-local. I'm not necessarily opposed to this, but I just want to understand the what and why.

README.md
```diff
@@ -272,6 +273,15 @@ The migration process currently consists of 3 commands:
 
 ---
 
+## Debugging with Pycharm
```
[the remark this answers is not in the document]
Niicck: It took me a couple hours to get all the configs and order of operations right for Pycharm to work for me. I'm documenting it for the sake of future devs/future me.

docker-compose.migration_test_override.yml
```diff
@@ -9,3 +9,5 @@ services:
     volumes:
       # Use local docker-entrypoint.sh instead of docker-entrypoint from older master image
       - "./docker-entrypoint.sh:/app/docker-entrypoint.sh"
+      # Use local scripts in case there were updates that are not in older master image
```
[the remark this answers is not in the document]
Niicck: Yes updates to the scripts. That line mounts your local scripts directory to the docker container's scripts directory, overwriting the old versions of scripts from the master branch. This is necessary when your local branch's scripts have important changes that have not been promoted to master yet.

scripts/serve-local.sh
Niicck: Undockerized Joplin doesn't have an entrypoint, so we need a script to run migrations and seed data. I updated setup-undockered-data to only load data if LOAD_DATA=on is passed, so it conforms to our existing flag setup.

scripts/sanitize_revision_data.sql
```diff
@@ -0,0 +1,14 @@
+-- wagtailcore_pagerevision.content_json.content_type uses foreign keys from production data.
+-- Foreign key ids inside of nested content_json field are not automatically updated during datadump.
```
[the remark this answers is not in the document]
Niicck: I don't think it is a bug. "content_json" is a dumb text column. It knows nothing about foreign key relationships, nor should it. Ideally, this problem should never exist. content_types would be generated in the same order on any environment, but our prod and dev migrations deviated at some point, which is our bad and not Wagtail's fault. I haven't seen anyone else have this same problem.

A contributed solution would entail either:
a) patching `manage.py loaddata` itself to apply my sql code to the "content_json" field, which seems aggressive/unfeasible.
b) fundamentally changing how revisions get saved and updated. Wagtail would have to know how and when to re-run updates to all past revisions. I think updating the sql at the instance the data is loaded is easier and more efficient.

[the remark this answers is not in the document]
easherma: Thanks for clarifying!

README.md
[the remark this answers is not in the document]
briaguya: I agree with @Niicck on this one, I went down a more complex path than this set of instructions when trying to get everything working with the pycharm debugger and I'm not sure what exactly was "magic" because I just happened to have my env set up in ways that worked and what was actual intentional.

If we want to change it to say, "Debugging with an IDE" and then have a subheading of "Pycharm" and a "TODO" for other IDEs that could work, but I'm very glad to see these instructions make it into the readme.

.circleci/config.yml
```diff
@@ -61,7 +61,6 @@ workflows:
           filters:
             branches:
               ignore:
-                - master
```
briaguya: I'm assuming this is so we can get review apps for our "Master -> Prod" PRs, @Niicck could you confirm?

Niicck: Nope. The stage name is now a little bit of a misnomer. I wanted to give us the ability to re-build/re-seed our staging database. I have discovered a problem with my approach. I will change it

docker-entrypoint.sh
```diff
@@ -36,6 +41,7 @@ case "${DEPLOYMENT_MODE}" in
     if [ "$LOAD_PROD_DATA" == "on" ]; then
       # Option for migration_test.sh to source directly from production
       python ./joplin/manage.py loaddata ./joplin/db/system-generated/tmp_production.datadump.json
+      sanitize_revision_data
```
easherma: Why only on Load_Prod_Data?

Is there an assumption that if you run with this flag locally once it will fix the problem moving forward?

Niicck: Load_Prod_Data is only run during migration-test, when your seeding data is initially created. So if you sanitize that data, then the datadump thats used as a future seeding.datadump.json will also be sanitized.

PR #309, opened 2019-07-27
2510 labels

easherma:
Fixes https://github.com/cityofaustin/techstack/issues/2510
We added the material icon font set, but it overrides the font family of the link label as well, and material icons does not have lowercase characters. 

So I extended the menu item template to add a span tag to give a font family to the label.


joplin/css/variables.scss
```diff
@@ -19,6 +19,6 @@ $coa-color-white: #FFFFFF;
 
 // LAYOUT
 
-$nav-wrapper-width: 80px;
+$nav-wrapper-width: 180px;
```
briaguya: 140 is nice and fits the stuffs

PR #310, opened 2019-08-02
Issue 2572: backend for official doc

Niicck:
Ref: cityofaustin/techstack#2572

Deployed: https://joplin-pr-2572-official-doc.herokuapp.com/admin/pages/3/

Test out making an official document list from the create content modal.


README.md
```diff
@@ -90,6 +90,8 @@ pipenv install --requirements deploy/requirements.txt
 sh scripts/undockered.sh
 ```
 
+Note! To run additional pipenv commands like `pipenv run ./joplin/manage.py migrate` you must have the variable `export DATABASE_URL=postgres://joplin@127.0.0.1:5433/joplin` in your bash scope.
```
easherma: interesting. what goes wrong if not? Couldn't we make this variable part of the undockered script?

joplin/base/models.py
```diff
@@ -696,6 +741,18 @@ class InformationPageRelatedDepartments(ClusterableModel):
...
 
+class OfficialDocumentPageRelatedDepartments(ClusterableModel):
+    page = ParentalKey(OfficialDocumentPage, related_name='related_departments', default=None)
+    related_department = models.ForeignKey(
+        "base.departmentPage",
+        on_delete=models.PROTECT,
+    )
+
+    panels = [
+        # Use a SnippetChooserPanel because blog.BlogAuthor is registered as a snippet
```
easherma: this looks like a bunk comment from a branch of mine that got merged without correcting

joplin/base/models.py
```diff
@@ -482,6 +481,52 @@ class DepartmentPageDirector(Orderable):
         FieldPanel('about'),
     ]
 
+class OfficialDocumentPage(JanisPage):
```
easherma: I know the current code base dosen't have this, but I'd love to have us get in the habit of using docstrings
Off the top of  my head: 
"""
This is a page for listing related documents for a department(s?) and arranging them by date.
It's currently essentially a list with a brief summary of each item and an link to the url of the document. 
Eventually this might migrate to/be replaced by a model using Wagtail Documents
"""

joplin/base/models.py
```diff
@@ -482,6 +481,52 @@ class DepartmentPageDirector(Orderable):
...
+    janis_url_page_type = "official_document"
+    base_form_class = custom_forms.OfficialDocumentPageForm
+
+    description = models.TextField(blank=True)
+
+    department = models.ForeignKey(
```
easherma: I could be misreading this but I think we need to decide whether to use the department model (one to many) or allow multiple related departments. 

Since you made a related departments clusterable model, I'd assume we want a document page to potentially belong to multiple departments(which sounds like an organizational mess but shrug emoji), so we might want to stick to one option or the other? 

Any thoughts @briaguya ?

joplin/base/templatetags/joplin_tags.py
```diff
@@ -25,6 +25,7 @@ def get_revision_preview_url(*args, **kwargs):
   'department page': 'pick-the-perfect-content-type/department-page',
   'topic page': '',
   'topic collection page': '',
+  'official document page': '',
```
easherma: Thanks for thinking of/including this :-D

joplin/base/models.py
```diff
@@ -482,6 +481,52 @@ class DepartmentPageDirector(Orderable):
...
+        FieldPanel('title_es'),
+        FieldPanel('title_ar'),
+        FieldPanel('title_vi'),
+        FieldPanel('description'),
+        InlinePanel('topics', label='Topics'),
+        InlinePanel('related_departments', label='Related Departments'),
+        InlinePanel('official_documents', label="Documents", heading="Entries will be listed by document date (newest first)."),
+    ]
+
+class OfficialDocumentPageOfficialDocument(Orderable):
```
easherma: This naming is confusing to me. Is this supposed to be referring to a particular document that you'd include on a 'list of documents' page?

A clarifying docstring would be great for when the code gets revisited later. Realizing this follows some existing naming conventions in the current codebase...

joplin/css/editor.scss
```diff
@@ -53,3 +53,9 @@ label {
         display: none;
     }
 }
+
```
easherma: Where did this come from/whats it for?

joplin/js/CreateContentModal/ChooseTitleStep.js
```diff
@@ -20,7 +20,7 @@ const ChooseTitleStep = ({
       )}
       {pageType !== 'department' && (
         <span>
-          Write an actionable title for your  {pageType} page, starting with a
+          Write an actionable title for your  {pageType.replace("_", " ")} page, starting with a
```
easherma: this feels like part of some tech debt, but I think it's good for now. 

Ideally in the future it would be great to find a way to pass model data to our modal, since Django already generates the model names and their verbose equivalents IIRC

joplin/js/editor.js
```diff
@@ -194,6 +196,7 @@ $(function() {
         }
       });
 
+    // Fields inside of InlinePanels
```
easherma: yes thank u for comments! <3

joplin/base/models.py
```diff
@@ -7,7 +7,7 @@
 
 from wagtail.utils.decorators import cached_classmethod
 from wagtail.admin.edit_handlers import FieldPanel, InlinePanel, MultiFieldPanel, ObjectList, StreamFieldPanel, TabbedInterface, HelpPanel
-from wagtail.core.blocks import TextBlock, RichTextBlock, ListBlock, StreamBlock, StructBlock, URLBlock, PageChooserBlock, CharBlock
+from wagtail.core.blocks import TextBlock, RichTextBlock, ListBlock, StreamBlock, StructBlock, URLBlock, PageChooserBlock, CharBlock, DateBlock
```
easherma: DateBlock is added here but I don't believe it is used?

joplin/api/schema.py
```diff
@@ -153,13 +159,39 @@ class Meta:
         filter_fields = ['id', 'slug', 'live']
         interfaces = [graphene.Node]
 
+class OfficialDocumentFilter(FilterSet):
```
easherma: siiick filter :-D

joplin/base/models.py
briaguya: multiple departments everywhere. I know we've done that for info/service pages, I figured we're keeping that pattern for documents pages too.

joplin/base/models.py
briaguya: I think current naming conventions are more along the lines of:
`ServicePage` - a page about a service
`Contact`- the contact info for something
`ServicePageContact` - The relationship between the ServicePage and the Contact

It seems as though this naming is more like
`OfficialDocumentPage` - the official document page
`OfficialDocumentPageOfficialDocument` - the documents that go on that page

Doesn't quite fit the (completely undocumented and not actually existing) naming conventions we have, but I see no issue with keeping it as is.

joplin/js/editor.js
```diff
@@ -202,9 +205,9 @@ $(function() {
         var langString = labelText.match(languageRegex);
 
         if (langString != null && langString != lowerLanguageStrings[currentLang]) {
-          elem.classList.add("hidden");
+          elem.parentElement.classList.add("hidden");
```
briaguya: ooh! is there an issue for the thing this fixes somewhere? do we get bonus magic?

README.md
Niicck: It is part of the undockered script. But if you want to just run manage.py commands outside of the script you have to include a DATABASE_URL. Otherwise it'll start building a sqlite database and running migrations on that.

joplin/css/editor.scss
Niicck: My IDE adds a newline at the end of every file I mess with.

joplin/js/editor.js
Niicck: Oh yeah bonus magic. I don't think it was even an issue yet

joplin/base/models.py
Niicck: Ah ok I see the naming pattern now.

I was going off this kind of pattern: `DepartmentPageDirector` is a Director for a Department Page. If many pages had Directors, then you'd want a standalone `Director` model and then a relationship which would be `DepartmentPageDirector`. But only `OfficialDocumentPages` will have `OfficialDocuments`. In case the word `Document` is later used to refer to built-in document uploading, I went with the more verbose `OfficialDocument`.

I'll add documentation.

joplin/base/models.py
Niicck: A fine idea. I will do this.

joplin/js/CreateContentModal/ChooseTitleStep.js
Niicck: I'd forgotten about verbose_name! That's probably what I'd want here.

joplin/base/models.py
easherma: OK, so we'll want to nix department being declared as a ForeignKey here then, yah?

joplin/js/CreateContentModal/ChooseTitleStep.js
easherma: For sure!
Just for ref: https://docs.djangoproject.com/en/2.2/ref/models/options/
"If this isn’t given, Django will use a munged version of the class name: CamelCase becomes camel case."

PR #312, opened 2019-08-02
upgrade

easherma:
offical upgrade

This should help with excessive warnings before leaving a page
Lots of accessibility tweaks, I'm not sure that we benefit from all of them without updating our extended templates


easherma:
Closed in favor of https://github.com/cityofaustin/joplin/pull/331

PR #316, opened 2019-08-06
2479-add-button-styling

easherma:
Fixes https://github.com/cityofaustin/techstack/issues/2479


joplin/base/wagtail_hooks.py
```diff
@@ -193,5 +198,32 @@ def joplin_page_listing_more_buttons(page, page_perms, is_parent=False):
             _('Delete'),
             reverse('wagtailadmin_pages:delete', args=[page.id]),
             attrs={'title': _("Delete page '{title}'").format(title=page.get_admin_display_title())},
-            priority=30
         )
+
+
+
+@hooks.register('register_rich_text_features')
+def register_help_text_feature(features):
```
briaguya: I'm looking forward to see what other things we find regarding draftail.

I remember looking for draft-js table extensions before going to markdown route, it might be worth checking out some of those now that we have a clean implementation of a custom button working.

joplin/base/wagtail_hooks.py
```diff
@@ -193,5 +198,32 @@ def joplin_page_listing_more_buttons(page, page_perms, is_parent=False):
...
+        'element': 'div',
+    }
+
+    features.register_editor_plugin(
+        'draftail', feature_name, draftail_features.BlockFeature(control)
+    )
+
+    features.register_converter_rule('contentstate', feature_name, {
+        'from_database_format': {'div.rich-text-button-link': BlockElementHandler(type_)},
+        'to_database_format': {'block_map': {type_: {'element': 'div', 'props': {'class': 'usa-button-primary rich-text-button-link'}}}},
```
briaguya: I think MVP requirements are button-primary only, do we know if we'll need to do secondary in the future? If so it'd be cool to make an issue for it.

easherma: IMO I think this is already identified as a pretty edge case (only one page that even needs it so far)

The path is clear and I think the code is obvious enough that we can revisit in the future, but we could add an additional code comment as a note to future self

joplin/base/wagtail_hooks.py
easherma: We've got this issue in the backlog: 
https://github.com/cityofaustin/techstack/issues/2333

Luckily Thibaud wrote draftail and we know him now, so we could do some research and talk shop about a table inside draftail. 

Another route there might be some combo of using a Table Block: https://docs.wagtail.io/en/v2.6.1/reference/contrib/table_block.html
Inside of a react-streamfield so we could control the presentation more.

PR #336, opened 2019-08-15
collect static into subfolders based on environment 

easherma:
Ok, pretty sure this is working as intended!

Locally: 
Locally we can't build straight to s3 at the moment, we would want to implement a way to set the environment variables. I think this would be a good idea because I think it's could to have consistent behavior between all environments, but for now running `collectstatic` locally just pops in your local folder. 
On a Review App:
This gets kinda confusing becuase we have multple layers of deployment. But TL;DR: 
* review app stuff is collected under `static/reviewapp`, 
* but you will find it in the `joplin-austin-gov-prs3` bucket, which is different from production or staging.
* you can test this(I did) by ssh-ing in or running heroku console and running the collectstatic command. (Note: I upped the verbosity for this command because it will skip if the files haven't been modified and that was tripping me up/is good to have a note of)
* browse to [https://joplin-pr-try-static-folders.herokuapp.com/admin/pages/106/edit/]( https://joplin-pr-try-static-folders.herokuapp.com/admin/pages/106/edit/) , inspect the imported css, and you'll see it comes from the right bucket and path.
*obvi next step for this would be to put it in a subfolder based on the name of each branch
On Staging or Production:
* atm the AWS bucket is hard-coded in Heroku! That's not a big deal, but it would probably be nice to modify it so that behavior is consistent. 
* that being said, I would expect collectstatic to deposit their css in static/staging and static/production respectively, based on how it is working for my review app. Testing it in staging would be the next logical step from here.

### DIFF
--- a/joplin/settings.py
+++ b/joplin/settings.py
@@ -128,6 +128,7 @@ DEPLOYMENT_MODE = os.environ.get('DEPLOYMENT_MODE', 'LOCAL')
 ISPRODUCTION = DEPLOYMENT_MODE == "PRODUCTION"
 ISSTAGING = DEPLOYMENT_MODE == "STAGING"
 ISREVIEWAPP = DEPLOYMENT_MODE == "REVIEW"
+ISLOCAL = DEPLOYMENT_MODE == "LOCAL"
 
 
 # Database
@@ -193,7 +194,7 @@ if ISPRODUCTION:
     STATIC_ROOT = os.path.join(BASE_DIR, 'static/production')
 if ISREVIEWAPP:
     STATIC_ROOT = os.path.join(BASE_DIR, 'static/reviewapp')
-else:
+if ISLOCAL:
     STATIC_ROOT = os.path.join(BASE_DIR, 'static/local')
 STATIC_URL = '/static/'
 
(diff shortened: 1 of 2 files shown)


joplin/settings.py
```diff
@@ -76,7 +77,8 @@
     'webpack_loader',
     'dbbackup',
     'smuggler',
-    'session_security'
+    'session_security',
+    'storages'
```
easherma: not 100% if this needs to be here or not tbh, examples online varied

joplin/settings.py
```diff
@@ -287,24 +297,31 @@
...
     # Specifying the location of files
-    STATICFILES_LOCATION = 'static'
+    if ISSTAGING:
+        STATICFILES_LOCATION = 'static/staging'
+    if ISPRODUCTION:
+        STATICFILES_LOCATION = 'static/production'
+    if ISREVIEWAPP:
+        STATICFILES_LOCATION = 'static/reviewapp'
+    if ISREVIEWAPP:
+        STATICFILES_LOCATION = 'static/local'
```
Niicck: So would all review apps share the same styles? If that's what this implementation demands, then maybe it would make sense to just have Joplin host the styles, rather than S3. That way the styles would be unique to each deployment.

easherma: This is not too different from how it’s currently set up, but the webpack bundle folder does have versioned hashes.

Also like I mentioned above, an obvious next step would be to make a subfolder based on the branch name or something like that.

PR #350, opened 2019-08-29
2856 url construction

easherma:
Related to https://github.com/cityofaustin/techstack/issues/2856

I found this actually had a affect on load time/performance

### DIFF
--- a/joplin/base/models/topic_page.py
+++ b/joplin/base/models/topic_page.py
@@ -63,6 +63,3 @@ class TopicPageTopicCollection(ClusterableModel):
     panels = [
         PageChooserPanel('topiccollection'),
     ]
-
-    def __str__(self):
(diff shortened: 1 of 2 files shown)


joplin/base/models/topic_page.py
```diff
@@ -63,6 +63,3 @@ class TopicPageTopicCollection(ClusterableModel):
     panels = [
         PageChooserPanel('topiccollection'),
     ]
-
-    def __str__(self):
-        return self.topiccollection.text
```
easherma: this field dosen't exist and makes checking relationships in the shell impossible

Niicck: Is this the only thing you changed? I can't tell with the linting. Can you turn off linting on your PRs until we get whole app on the same page?

PR #351, opened 2019-08-30
2856 and now for something completely different

easherma:
I missed Python too much so here's a stab at something more compact and a step towards something more abstract, given the fluctuating nature of our IA 

This supersedes https://github.com/cityofaustin/joplin/pull/350 
and https://github.com/cityofaustin/joplin/pull/349

### DIFF
--- a/joplin/base/models/janis_page.py
+++ b/joplin/base/models/janis_page.py
@@ -44,20 +44,24 @@ class JanisBasePage(Page):
 
     def janis_url(self):
         """
-        This function parses various attributes of content types to construct the
-        expected url structure for janis
+        This function parses various attributes of related content types to construct the
+        expected url structure for janis.
 
         For attributes with multiple relations, it ONLY takes the FIRST one.
         """
 
         try:
             """
-             these use ternary operators with some appropriat conditionals
+             These use ternary operators with some appropriate conditionals
              the idea is: return this value in these cases or tell use you got
-             nothing (see the privacy policy info page for example)
+             nothing (see the privacy policy info page for example).
 
-             TODO: make this more abstract, check if the order of conditionals
-             affects performance
+             'None' responses get filtered out and removed from the URL path.
+
+             TODO:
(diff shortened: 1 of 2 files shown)


joplin/base/models/janis_page.py
```diff
@@ -43,38 +43,84 @@ class JanisBasePage(Page):
...
+                topic_slug,
+                page_slug]
+            # join them together, filtering out empty ones
+            janis_url = '/'.join(filter(None, (paths_list)))
+            return janis_url
+        except Exception as e:
+            # right now this is a catch-all,
+            print("!janis url error!:", self.title, e)
+            return "#"
+            pass
```
Niicck: I like this solution a lot. I'm just wondering why you'd think an Exception would be thrown here? If anything, it looks like your solution is so complete, that we'd want the staging site to break and an Exception to be thrown if something catastrophically went wrong.

Niicck: Alternatively, if speed is a concern, should every content type have their own function to handle the construction of the janis_url? Then you could cut out a bunch of the conditional checks. Food for thought, not a mandate.

easherma: Thanks! These are good questions. 

1) Regarding staging, I'm 90% sure this function exists merely to provide correct preview links for janis from joplin. 

2) As written it does cover a variety of edge cases, so the catch-all is more of a guarantee and a place to start if a new edge case is found. Previously these edge cases (like privacy policy) or guides would essentially not work but not give any obvious indication of that. But the code was not safe from bugs either, as Desi found. 

3) The way Python evaluates try/except blocks means that there is no performance downside to having this here, but it does leave a (albeit kinda hacky) message in the log before passing. So I'm not sure there is much of a downside to keeping it. (The Exception class, btw, does ignore some system and user input exceptions so its pretty safe to use here: https://docs.python.org/3/library/exceptions.html#exception-hierarchy )

PR #354, opened 2019-08-30
2734-counts

easherma:
Fixes https://github.com/cityofaustin/techstack/issues/2734

Here's what I did:

- Took a look at the create content modal code to see how we were doing counts there
    - Didn't see anything super easy to re-use, though I know we've discussed wanting to refactor the modal
- Did some research, which I documented in the issue

Small Detour:
- I found this project https://pypi.org/project/django-countable-field/ which accomplished some of what I was looking for (simple API, easy way to reference as a widget inside Wagtail), with some bonus features (like the options for word or sentence count, being able to control direction of the count, etc)
- However, the project as is only had a widget for TextArea, and I wanted one for TextInput, so I forked it off (ta-da: https://github.com/cityofaustin/django-countable-field) and added a widget for TextInput
    - I used pipenv to edit and use this repo locally, but ran into an issue deploying as such:
    - It's pretty easy to install a dependency through pipenv or pip direclty from github, but the docker image we use for deploys does not have Git! We might want to change that eventually, but for now I learned how to publish and update the project on PyPi (https://pypi.org/project/coa-django-countable-field/ )

Back to Joplin:
- We're now using our forked version of that widget. I set up some initial defaults in models.widgets.py, and overrode them for specific use cases
- TODO; widgets.py is kinda a terrible name, maybe there is a better place for that stuff to live
- I would really like to consolidate some of our model info, there is a lot of repeated code. I will propose that in a different PR. 

Allowing negative counts:
- So this is part of the acceptance criteria
- TL;DR here is, I think, the best way to approach this:
  - We should approach limits as a author interface UX thing and not a technical thing
  - For one, Django (especially using Postgres) treats the max_length as essentially cosmetic; it is _not_ actually limiting the length of the field in the database for TextArea, and for TextInput widgets only if the underlying field is a CharField. There is not really much of a performance benefit for limiting fields.
    - But if you do set max_length in Django/Wagtail, the form input will literally not let you enter further characters past the limit
   - So I propose setting any max_length, if one is set at all, to be higher than the limit you might set with the widget. Better yet, just set the limit for that field using the widget so that it displays as intended. It sounds more complicated than it ends up being.


joplin/base/models/official_documents_page.py
```diff
@@ -55,10 +57,14 @@ class OfficialDocumentPageOfficialDocument(Orderable):
 
     panels = [
         FieldPanel('date'),
-        FieldPanel('title'),
-        FieldPanel('authoring_office'),
-        FieldPanel('summary'),
-        FieldPanel('name'),
+        FieldPanel('title', widget=countMe),
+        FieldPanel('authoring_office', widget=countMe),
+        FieldPanel('summary', widget=widgets.CountableWidget(attrs={
```
easherma: At first I wanted a cool and succinct way to override just the max_length of the default config I have set, but I don't think it's really worth that extra hassle.

joplin/base/models/widgets.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+GENERIC_TITLE_LIMIT = 58
+GENERIC_DESCRIPTION_LIMIT = 254
+
+AUTHOR_LIMITS = {
+
+    "title": GENERIC_TITLE_LIMIT,
+    "description": 254,
+    "additional_content": 5000,
```
easherma: Some of these limits feel somewhat arbitrary and I'll want to revisit with UX at some point. ``` additional_content ``` being so high sticks out to me, because if we are really allowing that much input, we might instead better serve users by having that be a rich text field. 

Also we now have the option to count by word, which I think might be better for authors in general for stuff like descriptions.

joplin/base/models/department_page.py
```diff
@@ -72,13 +75,17 @@ def __str__(self):
     base_form_class = DepartmentPageForm
 
     content_panels = [
-        FieldPanel('title_en'),
-        FieldPanel('title_es'),
+        FieldPanel('title_en', widget=countMe),
+        FieldPanel('title_es', widget=countMe),
         FieldPanel('title_ar'),
```
briaguya: while it would be nice to avoid needing to put all of these titles by language in here (I'm not sure if we have an issue on the board for cleaning this up), it'd be good to update the widget on all of these.

Also this raises the question: do we have different character limits for different languages?

easherma: We do not, but with this widget we can now make that more possible.

PR #355, opened 2019-09-04
2855 joplin pickers

easherma:
Fixes https://github.com/cityofaustin/techstack/issues/2855

Using some judicious template extension and hacking, we got the spec in place for a more consistent page chooser
I'm not suuuper happy with this for the long term, as it creates more of a load keeping templates up to date with wagtail changes, but it works for now and is a decent short/medium term solution.


joplin/templates/wagtailadmin/chooser/browse.html
```diff
@@ -0,0 +1,42 @@
+{% load i18n %}
+{% if page_types_restricted %}
+    {% if page_type_string == "base.departmentpage" %}
```
easherma: these conditionals are nested in a kinda messy way, but it does allow us to deduce what heading to have based on the context, which a fallback incase a picker gets used elsewhere

easherma:
viewable @ https://joplin-pr-2855-joplin-pickers.herokuapp.com/

joplin/templates/wagtailadmin/shared/header.html
```diff
@@ -0,0 +1,51 @@
...
+
+    icon - name of an icon to place against the title
+
+    tabbed - if true, add the classname 'tab-merged'
+    merged - if true, add the classname 'merged'
+
+    add_link - if present, display an 'add' button. This is a URL route name (taking no parameters) to be used as the link URL for the button
+    add_text - text for the 'add' button
+{% endcomment %}
+<header class="{% if merged %}merged{% endif %} {% if tabbed %}tab-merged{% endif %} {% if search_form %}hasform{% endif %}" role="banner">
```
briaguya: I see this is where hasform gets set, just curious as to when/why we're using it to set display:none

joplin/css/admin.scss
```diff
@@ -38,10 +38,29 @@
 
 .page-explorer header,
 .homepage header,
-.page-editor header {
+.page-editor header:not(.hasform) {
```
briaguya: I'm curious as to how the hasform logic works, it's not clear to me when/where/why it's being used.

joplin/templates/wagtailadmin/chooser/browse.html
```diff
@@ -0,0 +1,46 @@
+{% load i18n %}
+{% if page_types_restricted %}
```
briaguya: I'm not sure what page_types_restricted means, I'm thinking it means we set available page types in the model, which makes me curious as to some of the strings in this logic

joplin/templates/wagtailadmin/chooser/browse.html
```diff
@@ -0,0 +1,46 @@
...
+    {% elif page_type_string == "base.informationpage,base.servicepage" %}
+        {% trans "Select a page for the guide" as choose_str %}
+        {% include "wagtailadmin/shared/header.html" with title=choose_str  search_url="wagtailadmin_choose_page_search" query_parameters="page_type="|add:page_type_string icon="doc-empty-inverse" %}
+
+    {% else %}
+        {% trans "Choose" as choose_str %}
+        {% include "wagtailadmin/shared/header.html" with title=choose_str subtitle=page_type_names|join:", " search_url="wagtailadmin_choose_page_search" query_parameters="page_type="|add:page_type_string icon="doc-empty-inverse" %}
+
+    {% endif %}
+{% else %}
```
briaguya: just clarifying, this part means: if we don't have available page types set then use "choose a page" as our string.

joplin/templates/wagtailadmin/chooser/browse.html
```diff
@@ -0,0 +1,46 @@
...
+{% if page_types_restricted %}
+    {% if page_type_string == "base.departmentpage" %}
+    {% trans "Add a department that offers this page" as choose_str %}
+    {% include "wagtailadmin/shared/header.html" with title=choose_str  search_url="wagtailadmin_choose_page_search" query_parameters="page_type="|add:page_type_string icon="doc-empty-inverse" %}
+    {% elif page_type_string == "base.informationpage,base.servicepage" %}
+        {% trans "Select a page for the guide" as choose_str %}
+        {% include "wagtailadmin/shared/header.html" with title=choose_str  search_url="wagtailadmin_choose_page_search" query_parameters="page_type="|add:page_type_string icon="doc-empty-inverse" %}
+
+    {% else %}
+        {% trans "Choose" as choose_str %}
```
briaguya: I know this is a fallback and we don't have an example where this happens, but having simply "choose" nested deeper in the conditional than "choose a page" threw me off at first

joplin/templates/wagtailadmin/shared/header.html
```diff
@@ -0,0 +1,51 @@
...
+    {% block breadcrumb %}{% endblock %}
+    <div class="row nice-padding">
+        <div class="left">
+            <div class="col header-title">
+                <h1 class="icon icon-{{ icon }}">
+                {{ title }}{% if subtitle %} <span>{{ subtitle }}</span>{% endif %}</h1>
+            </div>
+            {% comment %}
```
briaguya: Any reason we have this commented out instead of just removing it completely?

easherma: I'm not 100% opposed to ditching it, I think its helpful to keep it as a reference for maintainability. Going back and forth between repos is kind of a pain, and its hard to see the diffs between vanilla wagtail and the changes we make and why.

joplin/templates/wagtailadmin/chooser/browse.html
easherma: Yep!

joplin/templates/wagtailadmin/chooser/browse.html
easherma: Hmm I'll try it out but that may have just been a typo on my end, but generally yes that is whats its used for (like for setting the help text appropriately below)

joplin/templates/wagtailadmin/shared/header.html
easherma: This is a vanilla wagtail thing (all we're doing is hiding the part thats commented), but I believe the submit is hidden because it autocomplets the search as you type, so having the submit button there would be kinda redundant, but the submit is still there for the form to work.  

Not 100% sure if we're talking about the same part though, happy to chat about it.

joplin/templates/wagtailadmin/chooser/browse.html
easherma: So this one renders as "Choose" {{ page_type_string}} which can be a list of types. It's generic but more specific than just Choose

joplin/css/admin.scss
easherma: so every page is a form, and modals can have forms too. For whatever reason in vanilla wagtail this attribute (.hasform) only pops up on modals, so its a good way to keep hiding headers elsewhere but un-hide them for modals. The logic of why or what else its used for I'm not sure about.

joplin/templates/wagtailadmin/chooser/browse.html
easherma: So I decided to leave this because its a way to avoid checking all these if statements if the picker isn't even restricted in the first place. Probably gives a minor performance benefit but I don't think it hurts either.

joplin/css/admin.scss
briaguya: is this the most specific selector we have available? if so, a comment along the lines of:
"we need to select on page chooser modals only, and this is the most specific selector I could find" would be more descriptive.

joplin/templates/wagtailadmin/chooser/browse.html
```diff
@@ -0,0 +1,49 @@
...
+    {% if page_type_string == "base.departmentpage" %}
+        {% trans "Add a department that offers this page" as choose_str %}
+    {% elif page_type_string == "base.topiccollectionpage" %}
+        {% trans "Select a topic collection" as choose_str %}
+    {% elif page_type_string == "base.topicpage" %}
+        {% trans "Select a topic " as choose_str %}
+    {% elif page_type_string == "base.informationpage,base.servicepage" %}
+        {% trans "Select a page for the guide" as choose_str %}
+    {% elif page_type_string == "base.informationpage,base.servicepage,base.guidepage,base.officialdocumentpage" %}
+        {% trans "Choose content to feature on the department page" as choose_str %}
```
briaguya: I understand why this exists this way now, but this logic will be messy when we need to have the same page_type_string for our picker on the editable home page.

We should find a better way to determine we're definitely "choosing content to feature on the department page"

easherma: I agree this is hacky and merely deductive. I could not find a better way other than using generic page chooser. The page type of the page you are on isn't passed to the template and there's little else to go off of for conditionals. 

Meanwhile the generic chooser makes it pretty easy to define custom page choosers with their own templates and titles, so we won't even need this nest of conditionals then.

briaguya: sounds like the issue to make is: "refactor custom page choosers to have their own templates and titles"

easherma: https://github.com/cityofaustin/techstack/issues/3016

PR #358, opened 2019-09-05
2743 remove external links

easherma:
Fixes https://github.com/cityofaustin/techstack/issues/2743

@briaguya do you know if there's any janis code or query stuff associated with this component at all?


briaguya:
@easherma re: anything in Janis: https://github.com/cityofaustin/janis/blob/master/src/js/queries/topicFragment.js looks like it doesn't mention external links at all, so I'm assuming we're fine if you've tried building janis against this Joplin branch

joplin/base/migrations/0084_remove_topicpage_external_services.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 2.2.4 on 2019-09-05 17:03
```
briaguya: The migrations here are confusing to me,

It seems as though the most recent migration should be this, as this is the code change that we're making.

Is there something strange about how merging migrations is working?

easherma: Its working as intended. Having all the migrations there would be more important if we had other branches depending on them. Since our branch structure is pretty simple and joplin is only being used by the alpha project atm it's probably a bit overkill.

joplin/base/migrations/0086_auto_20190910_1727.py
```diff
@@ -0,0 +1,38 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('base', '0085_merge_20190910_1727'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='officialdocumentpageofficialdocument',
```
briaguya: @easherma I guess my real question is how is this not already in master?

PR #360, opened 2019-09-06
2662 change options_description to rich text field

robbobfrh84:
This is the "Joplin" part of issue #2662. 
- Fixes: https://github.com/cityofaustin/techstack/issues/2662
- see: ["Janis" pull request](https://github.com/cityofaustin/janis/pull/501)
- Deployed at: https://joplin-pr-2662-service-step-st.herokuapp.com/admin/login/

This will make it so when creating a "step with options description" 
- [x] - Content creators will be able to have a rich text field for "Steps with Option"
- [x] - Specify placeholder text.

Screen Shot 2019-09-09 at 12 30 14 PM


desigonz:
@robbobfrh84 thanks for your great work on this! I conferred with Chase and Annie and we're changing course—we'd like to remove H1 styling for step with options. I'll also update this in #2662

**New acceptance criteria**
Janis
- [ ] Step with options text doesn't style as H1
Joplin
- [ ] When creating a step with option, the "options description" field is a rich text field

easherma:
@robbobfrh84 and like we discussed in the walkthrough, making ' Step with options text doesn't style as H1' correct will actually be a tweak to Janis and not Joplin!

Niicck:
Deployed at: https://joplin-pr-2662-service-step-st.herokuapp.com/admin/login/

joplin/js/CreateContentModal/richTextPlaceholder.js
```diff
@@ -0,0 +1,51 @@
...
+        }
+      }
+    }
+
+  }
+
+  function checkFlaggedAncestors(parent, placeholder){
+    // - Here, we check up through all the elements direct parent nodes (ancestors),
+    // - If a placeholder flag is present we add the class acordingly.
+    while (parent = parent.parentNode) {
```
Niicck: I think you can use jQuery here instead of manually traversing the DOM. See `insertWizardData.js`.

joplin/js/CreateContentModal/richTextPlaceholder.js
```diff
@@ -0,0 +1,51 @@
...
+      if (
+        parent.classList &&
+        parent.classList.contains(wagtailFlaggedClassname)
+      ) {
+        parent.classList.forEach( classname => {
+          const subClasses = classname.split('_');
+          if (subClasses[0] === "odd-value") {
+            placeholder.classList.add("odd-placeolder-elm")
+            placeholder.innerText = subClasses[1].replace(/-/g," ");
```
Niicck: I'm wondering why we need the logic of parsing the name of the subclass.
Could we simplify it to (in pseudocode)
```
if (className == ''odd-option-description")
then add inner-text="Option description" <-- hardcoded here
```
I'm not sure if its necessary to get the inner-text value from the next subclass. It's a little complicated and there's a potential for bugs later on if we insert other classes there. Let me know what you think.

joplin/js/CreateContentModal/richTextPlaceholder.js
```diff
@@ -0,0 +1,51 @@
+export default function (){
+  /* - MutationObserver for Wagtail JS handling
+    - This function will handle flagged elements for custom styling and behavior.
+    - Reference: https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver
+      - Includes option to stop observing.
+  */
+  const wagtailFlaggedClassname = 'odd-placeholder'
```
Niicck: Our undocumented convention is to prefix our classes with "coa-" (for City of Austin). Good thinking though!

joplin/css/editor.scss
```diff
@@ -65,3 +68,7 @@ label {
 .sequence-container.sequence-type-list {
     width: 100% !important;
 }
+
+.odd-placeolder-elm {
+  color: $coa-color-dark-gray2 !important;
```
Niicck: Is `!important` required?

joplin/js/editor.js
```diff
@@ -4,13 +4,15 @@ import '../css/preview.scss';
 import insertWizardData from './CreateContentModal/insertWizardData';
 import menuActiveState from './EditPage/menuActiveState';
 import toggleActivePanel from './SidebarPreview/toggleActivePanel';
+import richTextPlaceholder from './CreateContentModal/richTextPlaceholder';
```
Niicck: `CreateContentModal` is a specific piece of the application. When you click to add a new page in Joplin, a Modal pops up that walks you through that process. So everything related to the CreateContentModal lives in the CreateContentModal/ folder. Your code would fit better in the EditPage/ folder because it affects the edit page only.

joplin/js/CreateContentModal/richTextPlaceholder.js
```diff
@@ -0,0 +1,51 @@
...
+            placeholder.classList.add("odd-placeolder-elm")
+            placeholder.innerText = subClasses[1].replace(/-/g," ");
+          }
+        })
+      }
+    }
+  }
+
+  // Create an observer instance linked to the callback function
+  const observer = new MutationObserver(callback);
```
Niicck: I'm glad you discovered this feature! Good find!

joplin/js/CreateContentModal/richTextPlaceholder.js
robbobfrh84: coa! got it! I'll replace all my "odd"(s) with "coa"(s)🌆

joplin/css/editor.scss
robbobfrh84: I need to watch that! 🤦‍♂️Past experience created a bad habit there! AND, to mention/note-out why it's needed if so. 
- In this case it turns out it is required. Looks like Drafttail's editor classes render before our `editor.scss` style. At least that's my understanding.

joplin/js/CreateContentModal/richTextPlaceholder.js
robbobfrh84: Swapping it to a hard coded "Option Description" with one classname will simplifying & clarify it. I think you're right and it's just too convoluted to be add dynamic (and limited) text in wagtail to sneak through. I'll refactor for that👍

joplin/css/editor.scss
Niicck: All good then!

joplin/js/editor.js
robbobfrh84: good point! moved it there👍

joplin/js/CreateContentModal/richTextPlaceholder.js
robbobfrh84: So, I did look into this and the issue I was having was that I needed to only query on the direct parents, because I didn't want to change the placeholders for all the placeholders with our specified "coa"-class. I wasn't able to get jQuery to specify this, or would have to do another filter operation. 
- This manual approach just checks directly "up" (parent, grandparent... etc), avoiding conflicting "coa"-class elements.

Niicck: You are correct! Thanks for the explanation.

joplin/base/models/service_page.py
```diff
@@ -33,8 +33,12 @@ class ServicePage(JanisBasePage):
             )),
             ('step_with_options_accordian', StructBlock(
                 [
-                    ('options_description', TextBlock(
-                        'Describe the set of options')),
+                    ('options_description', RichTextBlock(
+                        features=WYSIWYG_SERVICE_STEP,
+                        # richTextPlaceholder.js searches for the class 'coa-option-description' and replaces placeholder text
```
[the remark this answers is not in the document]
easherma: I would also add a note that the text (""Option description") is currently hardcoded by richTextPlaceholder.js (i like to leave comments on tech debt because I feel like it makes intentions more obvious to someone who might come along later)

joplin/css/editor.scss
```diff
@@ -65,3 +68,7 @@ label {
 .sequence-container.sequence-type-list {
     width: 100% !important;
 }
+
+.coa-placeolder-elm {
```
easherma: should this read 'placeholder'?

PR #366, opened 2019-09-11
Master -> Prod for 2703 only

briaguya:
Verifying new functionality by updating snippets:
Screen Shot 2019-09-11 at 3 12 16 AM

Verifying previous functionality by updating pages of every type:
Screen Shot 2019-09-11 at 3 17 21 AM


joplin/base/signals.py
```diff
@@ -33,7 +33,7 @@ def generate_responsive_images(sender, **kwargs):
         image.get_rendition(f'width-{width}')
 
 
-def create_build_aws(instance, publish=False, request=None):
+def create_build_aws(content_type, instance, publish_action='edited', request=None):
```
Niicck: Is `content_type` only used for logging in our slack messages?

joplin/base/signals.py
```diff
@@ -187,15 +181,31 @@ def get_http_request():
         return None
 
 
+# By creating a signal reciever for each snippet model we have, we can avoid
+# needing to filter out large amounts of unwanted calls in our function logic
```
Niicck: I don't understand that, but it sounds like a good thing. Do you mean a "large amounts of unwanted [create_build_aws] calls" or something else?

joplin/base/signals.py
```diff
@@ -50,15 +50,9 @@ def create_build_aws(instance, publish=False, request=None):
     logger.debug("create_build_aws() Starting task")
 
     slack_message = ""
-    publish_action = ""
-
-    if (publish):
-        publish_action = "published"
-    else:
-        publish_action = "unpublished"
```
Niicck: Was this unused code?

joplin/base/signals.py
briaguya: yes, this is only for slack messages, figured it'd be more clear this way considering we have "Recycle and Reuse drop off center" as a map, location, and contact

joplin/base/signals.py
briaguya: this used to be used for our slack message (either say we published or unpublished a page). I changed it so we just pass in a publish_action as a string for publish/unpublish, and have the default just be "edited"

joplin/base/signals.py
briaguya: when I had `@receiver(post_save)` in there without a specified sender I was getting all sorts of silly post_save events from django_session etc.

While I'd like to be able to make it more generic (for all snippets post save do x) having "if you're not all sorts of other django post_save stuff" logic in the fn seemed like a bad plan

PR #371, opened 2019-09-17
adds additional url construction handiling

easherma:
Fixes https://github.com/cityofaustin/techstack/issues/2999

### DIFF
--- a/joplin/base/models/janis_page.py
+++ b/joplin/base/models/janis_page.py
@@ -1,6 +1,6 @@
 import os
 import graphene
-
+import traceback
 
 from django.db import models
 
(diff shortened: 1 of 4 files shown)


joplin/base/models/janis_page.py
```diff
@@ -106,6 +108,21 @@ def janis_url(self):
...
+                try:
+                    theme_slug = self.topics.first().topic.topiccollections.first().topiccollection.theme.slug or None
+                    topic_collection_slug = self.topics.first().topic.topiccollections.first().topiccollection.slug or None
+                except AttributeError as e:
+                    try:
+                        theme_slug = self.topiccollections.first().topiccollection.theme.slug or None
+                        topic_collection_slug = self.topiccollections.first().topiccollection.slug or None
+                    except AttributeError as e:
+                        # this is for pages just under departments
+                        theme_slug = self.related_departments.all()[0].related_department.slug or None
```
briaguya: not the biggest fan of having department named theme here, shouldn't the entire URL be /department_slug/page_slug in this scenario?

easherma: Thats the effective result. I agree that as is isn’t super legible, but abstractly ‘theme_slug’ is just the first path

PR #372, opened 2019-09-18
2862 global pages

briaguya:


joplin/base/models/janis_page.py
```diff
@@ -42,6 +42,8 @@ class JanisBasePage(Page):
         verbose_name='Notes for authors (Not visible on the resident facing site)'
     )
 
+    coa_global = models.BooleanField(default=False, verbose_name='Make this a top level page')
```
easherma: i think the verbose name needs to be updated (and im still sad we're abusing verbose name so much)

briaguya: That's one of the things I talked about needing to review with design/content.

briaguya: since we're just trying to get the privacy policy up asap, I've created https://github.com/cityofaustin/techstack/issues/3051 to handle this

PR #375, opened 2019-09-23
2827 edit page header

easherma:
Fixes https://github.com/cityofaustin/techstack/issues/2827

### DIFF
--- a/joplin/css/preview.scss
+++ b/joplin/css/preview.scss
@@ -1,6 +1,6 @@
 @import './variables';
 
-$flexcontainer-margin: 100px;
+$flexcontainer-margin: 135px;
 $preview-height: calc(100vh - #{$flexcontainer-margin});
 $fixedcontent-width: 375px;
 $sidebar-width: 377px;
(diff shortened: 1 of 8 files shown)


joplin/css/utilities/border-box-reset.scss
```diff
@@ -1,8 +1,8 @@
 // https://www.paulirish.com/2012/box-sizing-border-box-ftw/
 
-html {
-  box-sizing: border-box;
-}
-*, *:before, *:after {
-  box-sizing: inherit;
-}
+// html {
```
easherma: Mateo had this in here...I kinda get what its purpose is, but I don't know that we need it? I'm not a big fan of having a 'utility' that is essentially a global change, bound to have affects we won't think about as typically caused by this thing.

joplin/css/preview.scss
```diff
@@ -126,7 +126,7 @@ $sidebar-width: 377px;
     overflow: auto;
     right: 0px;
     box-shadow: 0 3px 3px 0 rgba(0, 0, 0, 0.2);
-
+    z-index: 45 !important;
```
easherma: TODO: z-indexes are kind of a mess

PR #376, opened 2019-09-24
2827 js refactor

easherma:
I condensed some of the language code down. It probably needs more work and testing, but the idea is not to have very similar code chunks for each particular admin UI object and just find all the language labels as they are and change them as needed.

### DIFF
--- a/joplin/js/editor.js
+++ b/joplin/js/editor.js
@@ -186,7 +186,8 @@ $(function() {
     }
     replaceLanguageLabels();
     var languageTage = null;
-    for (let [index, val] of Object.entries(state.languageLabels)) {
+    var labelList = state.languageLabels;
+    for (let [index, val] of Object.entries(labelList)) {
       if (labelList[index].querySelector) {
         var languageTag = labelList[index].querySelector('span').innerText;
         if (languageTag != null && languageTag != currentLang) {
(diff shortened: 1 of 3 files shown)


joplin/js/editor.js
```diff
@@ -88,7 +88,7 @@ $(function() {
       $(label).append(link);
     }
   }
-
+  // TODO: what is this for??
```
briaguya: Looks like it came from this: https://github.com/cityofaustin/joplin/commit/a067fb95ae0b00bf1e718503b02d068e02aa7a0d

joplin/js/editor.js
```diff
@@ -165,67 +166,58 @@ $(function() {
...
+          state.languageLabels = languageLabels;
+        } else {
+          for (let label in languageLabels) {
+            state.languageLabels.push(languageLabels[label]);
+          }
+        }
+        languageLabels.each(function() {
+          this.innerHTML = this.innerHTML.replace(
+            '[',
+            " <span style='display:none;'>",
```
briaguya: It might be nice to add a comment here to explain how this works, something like: "since our translation lib adds language strings we don't want added to our labels, let's hide them."

joplin/js/editor.js
```diff
@@ -272,6 +328,7 @@ $(function() {
     var observer = new MutationObserver(function(mutations, observer) {
       if (mutations[0].attributeName == 'value') {
         $(element).trigger('change');
+        console.log(mutations);
```
briaguya: quick note to remove console logs

PR #390, opened 2019-10-02
Fixed height fix for layout

briaguya:
based on this note: https://github.com/cityofaustin/techstack/issues/2827#issuecomment-537265221

"When page title breaks to two lines, the tab row still displays as expected"

This just used a fixed height for the header to avoid odd rearrangement issues. This should not be considered a permanent fix


easherma:
Is there a build up of this one? While this looks good to me at first glance, it might be good to have QA look and and thumbs up or down before merging to staging

PR #391, opened 2019-10-02
just check for lang in more spots

briaguya:
based on this note: https://github.com/cityofaustin/techstack/issues/2827#issuecomment-537265221

"Share button generates clipboard link as expected"

This seems to fix it, I'm not sure the logic is perfect.


easherma:
Can you share a bit about how you came up with this solution and bit about why this option worked/what you were able to deduce about the cause of the bug? That would be some helpful context to review this in.

PR #422, opened 2019-10-22
3285-error-messages

easherma:
# Description

This PR addresses some errors that were being printed but not handled. They are now handled as initially expected. 

It also has a little tweak to our build process to hopefully make it less sensitive to warnings being printed to the console in the future.


scripts/migration-test.sh
```diff
@@ -140,9 +140,8 @@ function handle_input {
     # Build new migration datadump
     # Excluding wagtailcore.GroupCollectionPermission because of IntegrityError issues
     # see: https://docs.djangoproject.com/en/2.2/topics/serialization/#natural-keys for details
-    docker exec -it ${COMPOSE_PROJECT_NAME}_app_1 python joplin/manage.py dumpdata --exclude=wagtailcore.GroupCollectionPermission --indent 2 --natural-foreign --natural-primary | \
-      python ./scripts/remove_logs_from_json_stream.py \
-      > ./joplin/db/system-generated/seeding.datadump.json
+    docker exec -it ${COMPOSE_PROJECT_NAME}_app_1 python joplin/manage.py dumpdata --exclude=wagtailcore.GroupCollectionPermission --indent 2 --natural-foreign --natural-primary \
```
easherma: dumpdata has an --output flag that, in this case, would avoid future problems with warnings printed to stout

joplin/base/models/janis_page.py
```diff
@@ -129,14 +129,23 @@ def janis_url(self):
                     except AttributeError as e:
                         # this is for pages just under departments
                         theme_slug = self.related_departments.all()[0].related_department.slug or None
-
+                    finally:
```
easherma: def some tech debt here; repeating the same code twice is ugly, but it needed to be under this separate conditional in order for all use cases to be covered. I tried writing a function to call that would encapsulate this, but it didn't really make much sense. Open to ideas/future refactoring

Niicck: Why is copy + pasting into "finally" required? Wouldn't the code just keep going and create the paths list?

The double nested Attribution Error catching seems weird. Is there a clearer way to express what's happening? Like, should each Attribution Error only happen for a particular content type?

Niicck: ^ 2nd part isn't part of this issue, feel free to ignore.

easherma: @Niicck actually I think the 2nd is related to first. I must have assumed it would keep going when I first wrote it, but it dosen't. Probably because they are nested inside that 'if' statement

easherma:
> Couple questions and comments.
> 
> Also, I'm not sure what problem this is supposed to solve. There's no info in the PR documentation and the issue it references doesn't exist.

@Niicck I connected the PR to it's appropriate issue and added a description.

PR #425, opened 2019-10-22
Wagtail 27 test

easherma:
The future is now


joplin/js/editor.js
```diff
@@ -214,13 +214,28 @@ $(function() {
...
           if (languageTag != null && languageTag != currentLang) {
             translatedElement.classList.add('hidden');
           } else {
             translatedElement.classList.remove('hidden');
           }
+          /*  While the first condition checks for 'struct-blocks' with language tags,
+              it doesn't catch the case where there are 'struct-blocks' with language
+              tages WITHIN the element itself. The following condition checks for those conditions.
+              - The was neccessary for guide stream fields. */
+          if (translatedElement.classList.contains("struct-block")) {
```
briaguya: is there any chance we have a null/blank translatedElement that could cause an error to pop up here?

easherma: I don't think so. The main way I'd see that happen is if there is something totally messed up with the page render. In any case, we'd have plenty to solve then. I think there is probably more refactoring that could be done in the future, but I don't think this is any more at risk of that sort of bug than the old code.

PR #439, opened 2019-10-30
2581 tweaks

easherma:
# Description

1) The first commit changes where we exclude home and root, avoiding an extra for loop and making the query about 50% faster
2) I included the search input that works using AJAX. See comments on the PR. 
3) I added back the old header but in a more flexible way, cause it still may show up on some views. 

There were some other issues Bob and I discussed and raised, but we can turn those into follow up issues.

### DIFF
--- a/joplin/base/views/joplin_search_views.py
+++ b/joplin/base/views/joplin_search_views.py
@@ -12,6 +12,11 @@ from wagtail.admin.forms.search import SearchForm
  - See referenced file here: https://github.com/wagtail/wagtail/blob/a459e91692659aba04e662978857d14061aecaee/wagtail/admin/views/pages.py#L917
  - Joplin needs more control over our searches in how we sort/filter and display
  the page data the out of the box wagtail provides.
+ The main things that we changed from vanilla:
+ - we 'run' the query as soon as you visit the page (before any search),
+   so we can populate it with summary counts and sort stuff
+ - excluding certain pages from the pages to query
+
 """
 
 
@@ -68,11 +73,6 @@ def search(request):
 
     else:
         form = SearchForm()
-        """
-          JOPLIN NOTE:
-          - This is where we 'hide' the 'home' and 'root' page on initial load of main content page.
-          - However, these pages will be available in any search that matches title.
-        """
 
     # "Content Type Builder" Joplin Note: Moved from query condition above.
     all_pages = all_pages.search(q, order_by_relevance=not ordering, operator='and')
--- a/joplin/base/views/joplin_search_views.py
+++ b/joplin/base/views/joplin_search_views.py
@@ -14,8 +14,10 @@ from wagtail.admin.forms.search import SearchForm
  the page data the out of the box wagtail provides.
 """
 
+
 def search(request):
-    pages = all_pages = Page.objects.all().prefetch_related('content_type').specific()
+    # excluding Root(1) and Home(3) pages from search
+    pages = all_pages = Page.objects.all().exclude(id__in=[1, 3]).prefetch_related('content_type').specific()
     q = MATCH_ALL
     content_types = []
     pagination_query_params = QueryDict({}, mutable=True)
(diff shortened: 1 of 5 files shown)


joplin/settings.py
```diff
@@ -242,6 +242,7 @@
         'debug_toolbar.panels.headers.HeadersPanel',
         'debug_toolbar.panels.request.RequestPanel',
         'debug_toolbar.panels.sql.SQLPanel',
+        'debug_toolbar.panels.templates.TemplatesPanel',
```
easherma: this is just a handy thing to have, adding it back to debug toolbar

joplin/templates/wagtailadmin/pages/listing/_table_headers_search.html
```diff
@@ -0,0 +1,81 @@
+{% load i18n %}
```
easherma: we arent using this at all, i just stubbed it hear to stash the changes @robbobfrh84 made. 

I'm not super clear on why or if they are needed, but if they are, I would like to have a new template partial specifically for search or otherwise differentiate between other times the explore template might get used.

joplin/templates/wagtailadmin/pages/listing/_table_headers_search.html
```diff
@@ -0,0 +1,81 @@
...
+{% endcomment %}
+
+<tr class="table-headers">
+    {% if show_ordering_column %}
+        <th class="ord">
+            {% if orderable %}
+                {% if ordering == "ord" %}
+                    <a href="?{{ request.GET.urlencode }}" class="icon icon-order text-replace" title="{% trans 'Disable ordering of child pages' %}">{% trans 'Order' %}</a>
+                {% else %}
+                    <a href="?{{ request.GET.urlencode }}&ordering=ord" class="icon icon-order text-replace" title="{% trans 'Enable ordering of child pages' %}">{% trans 'Order' %}</a>
```
easherma: btw @robbobfrh84 , i think this is where your sorting queries kept getting tagged on

joplin/templates/wagtailadmin/pages/coa_header.html
```diff
@@ -0,0 +1,21 @@
+{% load wagtailadmin_tags staticfiles i18n %}
```
easherma: adding this back cause the header is still potentially used on some pages and will look weird without them

joplin/templates/wagtailadmin/pages/index.html
```diff
@@ -0,0 +1,7 @@
+{% extends "wagtailadmin/pages/index.html" %}
+{% load wagtailadmin_tags staticfiles i18n %}
+
+{% block content %}
+{% include "wagtailadmin/pages/coa_header.html" %}
```
easherma: probably would be a good place to include different headers depending on params (like 'are we searching?')

joplin/templates/wagtailadmin/pages/search_header.html
```diff
@@ -10,13 +10,23 @@ <h1 class="coa-search-results-h1">
 
     <div>
         <!-- From wagtail/admin/templates/wagtailadmin/shared/menu_search.html -->
-        <form class="nav-search" action="{% url 'wagtailadmin_pages:search' %}" method="get">
+        {% if search_url %}
+            <form class="col search-form nav-search" action="{% url search_url %}{% if query_parameters %}?{{ query_parameters }}{% endif %}" method="get" novalidate>
```
easherma: this form does ajax magic. i still don't fully understand why. but i know its one of the main reasons why we wanted to move to search page

joplin/base/views/joplin_search_views.py
```diff
@@ -12,10 +12,17 @@
...
+   so we can populate it with summary counts and sort stuff
+ - excluding certain pages from the pages to query
+
 """
 
+
 def search(request):
-    pages = all_pages = Page.objects.all().prefetch_related('content_type').specific()
+    # excluding Root(1) and Home(3) pages from search
+    pages = all_pages = Page.objects.all().exclude(id__in=[1, 3]).prefetch_related('content_type').specific()
```
easherma: much faster than looping through titles

PR #448, opened 2019-11-04
3363 Main search alert & styling fixes

robbobfrh84:
# Description

Issue: https://github.com/cityofaustin/techstack/issues/3363

Styling 

* [x] Styling for search box isn't centered. 
* [x] The header is covering part of the content_types as a transparent padding. 
* [x] Make Search Icon and font color fixed until design can give feedback/set style

Click search box after typing
Initial

Behavior Issues

* [x] No search + content_type = "This field is required." error
* [x] Search by Autocompolete (don't hit enter) click content_type filter, see error.
  * "Changes you made may not be saved"

Search style fix

# Testing Notes

... Search for something by typing into search box 
- the page should update while you type (no need to hit enter or select a "search" icon.

Now, with the search results return, filter by content types
- The content type should now be filtered without an error message or alert. 

* [Deployed Link](https://joplin-pr-search-fixes.herokuapp.com/admin/pages/search/)

# Checklist:
- [x] Request reviewers

### DIFF
--- a/joplin/base/views/joplin_search_views.py
+++ b/joplin/base/views/joplin_search_views.py
@@ -94,7 +94,7 @@ def search(request):
         return render(request, "wagtailadmin/pages/search_results.html", {
             'pages': pages,
             'all_pages': all_pages,
-            'query_string': q,
+            'query_string': query,
             'content_types': content_types,
             'selected_content_type': selected_content_type,
             'ordering': ordering,
--- a/joplin/css/layouts/headers.scss
+++ b/joplin/css/layouts/headers.scss
@@ -48,9 +48,24 @@
   height: 40px !important;
 }
 
-.coa-nav-search li {
-  margin: 0px !important;
-  padding: 0px !important;
+.nav-search li {
+  margin: 0px;
+  padding: 0px;
+  list-style-type: none;
+}
+
+.icon-search:before {
+  margin-top: 5px !important;
+  color: #777 !important;
+}
+
+.icon-search input {
+  color: #262626 !important;
+  background-color: hsla(0,0%,39%,.15) !important;
+}
+
+.icon-search input::placeholder {
+  color: #aaa !important;
 }
 
 .listing-filter {
--- a/joplin/css/layouts/headers.scss
+++ b/joplin/css/layouts/headers.scss
@@ -48,6 +48,11 @@
   height: 40px !important;
 }
 
+.coa-nav-search li {
+  margin: 0px !important;
+  padding: 0px !important;
+}
+
 .listing-filter {
   margin: 0 !important;
   padding: 20px !important;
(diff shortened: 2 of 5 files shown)


joplin/base/views/joplin_search_views.py
```diff
@@ -94,7 +94,7 @@ def search(request):
         return render(request, "wagtailadmin/pages/search_results.html", {
             'pages': pages,
             'all_pages': all_pages,
-            'query_string': q,
+            'query_string': query,
             'content_types': content_types,
```
robbobfrh84: The 'q' value does other things and we just want to pass the actual query,  it was causing an error when pages rendered without an actual search.

joplin/css/layouts/headers.scss
```diff
@@ -48,6 +48,26 @@
...
+}
+
+.icon-search input {
+  color: #262626 !important;
+  background-color: hsla(0,0%,39%,.15) !important;
+}
+
+.icon-search input::placeholder {
+  color: #aaa !important;
+}
```
robbobfrh84: these styles are a rough pass at making the input state static for "inside" the text box or not. We can tackle how we handle /design soon!

joplin/templates/wagtailadmin/pages/search_header.html
```diff
@@ -11,22 +11,12 @@ <h1 class="coa-search-results-h1">
...
-        {% endif %}
-        <!-- <form class="nav-search" action="{% url 'wagtailadmin_pages:search' %}" method="get">
-            <div>
-                <label for="menu-search-q">{% trans "Search" %}</label>
-                <input type="text" id="menu-search-q" name="q" placeholder="{% trans 'Search' %}" />
-                <button class="button" type="submit">{% trans "Search" %}</button>
+            <div class="col search-form nav-search">
+                {% for field in search_form %}
+                    {% include "wagtailadmin/shared/field_as_li.html" with field=field field_classes="field-small iconfield" input_classes="icon-search" %}
+                {% endfor %}
```
robbobfrh84: Here is the root of where the issue was... 
- the "action" was redundant, as the imbedded ajax request was already doing the querying for us.
- So removing the actual html form behavior just let the input be an input, and, filtering now doesn't have to care what the input state of that form is.
- Before it freaked out when it saw new values in the search box and did and alert("you have unsaved data")

joplin/templates/wagtailadmin/shared/field.html
```diff
@@ -0,0 +1,21 @@
+<!--
+Joplin edited from Original file: https://github.com/wagtail/wagtail/blob/master/wagtail/admin/templates/wagtailadmin/shared/field.html
+-->
```
robbobfrh84: We just need this template so that we can remove the build in error messages.

joplin/templates/wagtailadmin/pages/search_header.html
chiaberry: ah, so it was trying to do too much?

joplin/templates/wagtailadmin/shared/field.html
chiaberry: thanks for including a link to the original version

PR #449, opened 2019-11-06
3146 - get unpublish back on the edit page

briaguya:
# Description

turns out we just had some dom issues with the dropdown

Fixes # (issue)

https://github.com/cityofaustin/techstack/issues/3146

# Testing Notes

If we can see the unpublish button on the edit page we're good

* [Deployed Link](https://joplin-pr-3146-edit-page-unpub.herokuapp.com/)

# Checklist:
- [x] Request reviewers

### DIFF
--- a/joplin/templates/wagtailadmin/pages/edit.html
+++ b/joplin/templates/wagtailadmin/pages/edit.html
@@ -106,7 +106,9 @@
                 <ul role="menu">
                   {# Save Draft should be secondary action if page is live #}
                   {% if page.live %}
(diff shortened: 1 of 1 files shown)


easherma:
This is about to be the 'wrong' way to do this: 
https://docs.wagtail.io/en/v2.7/releases/2.7.html#template-change-for-page-action-menu-hooks

chiaberry:
> This is about to be the 'wrong' way to do this:
> https://docs.wagtail.io/en/v2.7/releases/2.7.html#template-change-for-page-action-menu-hooks

when are we upgrading to 2.7? are there any breaking changes.

PR #464, opened 2019-11-19
3296 location mvp mvp

easherma:
# Description

By popular demand! A naive implementation of location page, with lots of copy pasta, and a copy pasta factory to make you more pasta.

# Types of changes

# How can this be tested?

 

# Checklist:
- [ ] Request reviewers


joplin/base/templatetags/joplin_tags.py
```diff
@@ -32,7 +34,11 @@ def get_revision_preview_url(*args, **kwargs):
 @register.simple_tag
 def get_style_guide_url(*args, **kwargs):
     content_type = kwargs['content_type'].name
-    return os.environ['STYLEGUIDE_URL'] + STYLEGUIDE_PAGES[content_type]
+    try:
+        get_style_guide_url = os.environ['STYLEGUIDE_URL'] + STYLEGUIDE_PAGES[content_type]
+    except Exception as e:
```
easherma: gives a default fallback for a page if its missing a styleguide url

joplin/base/models/janis_page.py
```diff
@@ -165,11 +164,14 @@ def janis_preview_url_end(self, revision=None):
         """
         if revision:
             url_page_type = revision.page.janis_url_page_type
+            global_id = graphene.Node.to_global_id('PageRevisionNode', revision.id)
```
easherma: @Niicck ran into some errors when making LocationPage a JanisPage (where I think there werent any revisions yet?)
anyways, this seems to do what you were looking to do while avoiding triggering an error if global_id fails for not existing

joplin/base/wagtail_hooks.py
```diff
@@ -58,7 +58,7 @@ def configure_main_menu(request, menu_items):
     menu_items[:] = [item for item in menu_items if item.name not in
                      # here were excluding some default generated menu items per UX
                      [
-                         'explorer',
+                         # 'explorer',
```
easherma: we can re-hide this later, atm I'm using it as an interface to show how index pages might work. in another branch i have an example using ModelAdmin, but thats probably overkill for this situation

joplin/base/models/janis_page.py
Niicck: Ah, nope different logic going on.
```
Optional "revision" parameter to get the janis_preview_url for a specific revision
Otherwise, it will return the janis_preview_url for the latest revision
```
Now you've changed it so that the default preview for a page won't preview the latest revision.

This logic will have to be changed back. If there are errors getting a preview_url for LocationPage that doesn't have a revision, it needs to be handled in a different way.

joplin/base/templatetags/joplin_tags.py
Niicck: You mean if its missing a STYLEGUIDE_PAGES[content_type] ?

joplin/locations/fixtures.py
```diff
@@ -0,0 +1,62 @@
...
+        this is probably a bad way to do this, we don't want to accidentially delete
+        data we care about, right?
+        """
+        try:
+            import locations
+            locations.models.LocationPage.objects.all().delete()
+        except Exception as e:
+            print(e)
+            import pdb
+            pdb.set_trace()
```
Niicck: We want to take out pdb.set_trace() in the Exception, even for mvp-mvp.

Is there more specific cleanup/handling to be done if an exception is caught here?

And why is this probably a bad way to handle clean up?

easherma: I think its 'probably' bad becuase its buried in a script that might get called in different contexts but it deletes all pages of that type in the database

Sooo say you have data on your branch, and then want to call this script to add test data. It would make test data on top of your data, than delete ALL data for the page type. That seems no bueno to me.

joplin/base/models/janis_page.py
easherma: Ok, my bad. Good to know. I'll revisit this.

joplin/base/templatetags/joplin_tags.py
easherma: Yes

joplin/base/models/janis_page.py
easherma: @Niicck this should be correct now

joplin/base/models/janis_page.py
```diff
@@ -257,6 +260,9 @@ def get_edit_handler(cls):
     class Meta:
         abstract = True
 
+# TODO: tech debt here, added an additional rule to hide this for location page
```
chiaberry: Where did you add the additional rule?

also my branch (that I just merged to master) has some changes here, maybe that can be extended to hide for location

joplin/js/editor.js
```diff
@@ -387,4 +396,36 @@ $(function() {
...
+
+  // autocomplete for places
+  (function() {
+    var latlng = {
+      lat: 30.26,
+      lng: -97.73,
+    };
+    var placesAutocomplete = places({
+      appId: 'plHX3G9C5GRN',
+      apiKey: 'd9275f5f90784eb5ba59ff8ec581d9bb',
```
chiaberry: should we hide these keys?

joplin/js/editor.js
```diff
@@ -387,4 +396,36 @@ $(function() {
...
+      container: document.querySelector('#id_physical_street'),
+      templates: {
+        value: function(suggestion) {
+          return suggestion.name;
+        },
+      },
+    }).configure({
+      aroundLatLng: latlng.lat + ',' + latlng.lng, // Paris latitude longitude
```
chiaberry: paris? its a leftover comment, right?

joplin/js/editor.js
easherma: we could, or import them as an env variable. but a)it was something slapped together as a quick POC, I haven't confirmed that we even want to use this particular service and b) im pretty sure the key would be exposed client side either way, i dont think its a real high priority at the moment but when we refactor we could address it

joplin/js/editor.js
easherma: Oui, good catch!

Pipfile
```diff
@@ -30,6 +30,7 @@ phonenumberslite = "==8.10.16"
 django-debug-toolbar = "==2.0"
 django-countable-field = {editable = true,git = "https://github.com/cityofaustin/django-countable-field"}
 django-flags = "*"
+wagtail-factories = "*"
```
Niicck: Can you add all the dependencies you're going to need to the Pipfile as well? I've got a change coming that's going to use Pipfile as the single source of truth. And maybe things like Faker can go into dev-dependencies.

joplin/base/models/janis_page.py
```diff
@@ -168,8 +167,12 @@ def janis_preview_url_end(self, revision=None):
         else:
             revision = self.get_latest_revision()
             url_page_type = self.janis_url_page_type
-        global_id = graphene.Node.to_global_id('PageRevisionNode', revision.id)
-        url_end = f"preview/{url_page_type}/{global_id}"
+        try:
+            global_id = graphene.Node.to_global_id('PageRevisionNode', revision.id)
+            url_end = f"preview/{url_page_type}/{global_id}"
+        except AttributeError:
+            url_end = f"preview/{url_page_type}/"
```
Niicck: Sorry I forgot, when does this happen again? If a page doesn't have a revision yet? Do locations not have revisions?

If this exception has been added just for the sake of MVP locations, let's make a "TODO" note to clean up previews on locations and then get rid of this logic. Because a preview of a page that has no revision should never exist.

Pipfile
easherma: this is the only new dependency I added, it uses Faker as a dependency directly 

All the changes in requirements.txt are becuase I re-froze it, and pip sorts requirements.txt alphabetically by default 

FYI @briaguya this addresses your question about requirements.txt as well

joplin/base/models/janis_page.py
easherma: IIRC it had to do with loading fixtures or generating test pages, those pages exist but I think they didn't have revisions associated with them yet. So it could happen with any page that is created but not with an associated revision. 

Example test page:
https://joplin-pr-3296-location-mvp-mv.herokuapp.com/admin/pages/267/revisions/

A location page has revisions like any other Wagtail Page type. 

Since we're writing our own url logic here, I think it's our responsibility to handle edge cases. In this case I wouldn't expect this to ever be a problem in staging or prod unless we progamatically enter content at some point(which could very well happen), but the try statement here has no performance drawbacks but helps avoid unexpected problems that might break the admin. 

From my perspective, good candidates for refactoring in this case would be how we handle previews (using https://github.com/torchbox/wagtail-headless-preview would be nice) and re-tooling to leverage wagtails existing functions for building urls which are more robust (or least informing our design based on them: 
https://github.com/wagtail/wagtail/blob/8238a338888258554ed6eae0e25eaf903a0a5831/wagtail/core/models.py#L748 )

Niicck: But this doesn't handle the edge case. It's just going to allow previews to break for test fixtures.

If fixtures don't have revision data, that should be accounted for in logic farther up the line rather than returning a preview_url that won't work.

But for the sake of getting the locations MVP merged in, let's worry about that later.

easherma: Yes, to be clear, it doesn't make previews work(which I didn't really have in scope for MVP as far as I'm aware), but without this try statement the entire build was breaking.

PR #465, opened 2019-11-20
3396 dummy data

briaguya:
This is just to get us a start on having a way to manage dummy data dumps. The included datadump has one of each page type, with very minimal data. Janis master will build without errors using the included data, and all pages are published/linked in nav.

### DIFF
--- a/.circleci/scripts/build_image.sh
+++ b/.circleci/scripts/build_image.sh
@@ -21,7 +21,7 @@ log 2 "Application Name:  ${APPNAME}"
 # Builds using top-level directory ($CURRENT_DIR/..) as context
 DOCKER_BUILDKIT=1 docker build -f app.Dockerfile -t $DOCKER_TAG_1 -t $DOCKER_TAG_2 -t $DOCKER_TAG_HEROKU --target $DOCKER_TARGET $CURRENT_DIR/../..
 
-if [ "$CIRCLE_BRANCH" == "master" ] || [ "$CIRCLE_BRANCH" == "production" ] || [ "$CIRCLE_BRANCH" == "3396-dummy-data" ]; then
+if [ "$CIRCLE_BRANCH" == "master" ] || [ "$CIRCLE_BRANCH" == "production" ]; then
   # Push master and production images to dockerhub repo for storage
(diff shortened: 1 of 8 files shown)


.circleci/scripts/build_image.sh
```diff
@@ -21,7 +21,7 @@ log 2 "Application Name:  ${APPNAME}"
 # Builds using top-level directory ($CURRENT_DIR/..) as context
 DOCKER_BUILDKIT=1 docker build -f app.Dockerfile -t $DOCKER_TAG_1 -t $DOCKER_TAG_2 -t $DOCKER_TAG_HEROKU --target $DOCKER_TARGET $CURRENT_DIR/../..
 
-if [ "$CIRCLE_BRANCH" == "master" ] || [ "$CIRCLE_BRANCH" == "production" ]; then
+if [ "$CIRCLE_BRANCH" == "master" ] || [ "$CIRCLE_BRANCH" == "production" ] || [ "$CIRCLE_BRANCH" == "3396-dummy-data" ]; then
```
briaguya: the plan is to remove this once we get into staging, we don't need extra dockerhub tags for this issue once it's merged in

scripts/migration-test.sh
```diff
@@ -54,6 +54,10 @@ elif [ "$LOAD_STAGING_DATA" = "on" ]; then
     python ./scripts/remove_logs_from_json_stream.py | \
     jq '(.[] | select(.model == "users.user") | .fields.password) |= "pbkdf2_sha256$150000$GJQ1UoZlgrC4$Ir0Uww/i9f2VKzHznU4B1uaHbdCxRnZ69w12cIvxWP0="' \
     > $TMP_DATADUMP
+elif [ "$DUMMY" = "on" ]; then
+  export DOCKER_TAG_APP="cityofaustin/joplin-app:pr-3396-dummy-data-latest"
```
briaguya: once this gets into staging this will be changed to use master instead

easherma:
Part of https://github.com/cityofaustin/joplin/pull/464 has a roadmap for how we could generate dummy data to then use in the dummy data dump

PR #467, opened 2019-11-20
awkward but functional implementation of django-recurrence 

easherma:


easherma:
this was a demo for @toribr ...I don't think we're using it for now, though the models in the plugin will probably be handy when we want to flesh out more elaborate time stuff (especially since they are rfc2445 compliant aka an accepted structure for icalender exports etc)

PR #507, opened 2019-12-22
3313 Limit "Make top-level link" to service and info pages

chiaberry:
# Description

Users should only be allowed to mark Service Pages and Information Pages as top level links. So in the same place where I check if a user is an admin, I added a check to see if it the correct pages before returning the check box and text. 

# Types of changes

# How can this be tested?

 

# Checklist:
- [ ] Request reviewers


joplin/js/CreateContentModal/ChooseTitleStep.js
```diff
@@ -83,8 +83,7 @@ const ChooseTitleStep = ({
             Use words you think residents may search to find the {pageType}.
           </li>
           <li>
-            You don’t need to worry about including your department’s name in
-            the title.
+            You don’t need to include your department’s name in the title.
```
chiaberry: this is the microcopy change for issue #3444 (replace joplin microcopy for content types)

chiaberry: I spun this out into its own PR and branch now

chiaberry:
top level links are indeed different from top pages on a topic page. Eric made a comment on the issue that "We should clarify the naming here, easy to mix up these 'top level link' checkbox with 'top links' on a topic page.

Also there is almost certainly a better way to designate these, but that might be more of a Joplin 3.0 thing."

PR #529, opened 2020-01-15
Async and pool

easherma:
# Description

Main things that happened here:

Switching gunicorn to run async workers. 

Doing this causes a cascade effect where we run out of connections in heroku becuase Django will set up a db connection for each request. 

Connection pooling would address this. 

The go-to way to do this is normally with pgbouncer, but I took @Niicck 's advice and looked for some ways to do this at the django app level instead of the db level. I tried https://github.com/jneight/django-db-geventpool , but it didn't seem to work correctly. 

I settled on https://github.com/lcd1232/django-postgrespool2 , which I don't 'like' as much, but actually seemed to 'work' (note that testing these things is a huge PITA as its hard to isolate variables, etc). 

I keep attempting to write docs re performance and timeout problems and running into things to try to just address the issue directly. 

So one big problem/difference with performance on builds vs using the Joplin app is that there is a really long queuing time for requests. I think that this is partially because of our inefficient API query, but _also_ the nature of the react-static build process. It is spinning out a number of requests for data all at once (I'm assuming). Gunicorn's workers are, by default, designed to be synchronous:
http://docs.gunicorn.org/en/stable/design.html

Note:
>This resource bound assumption is why we require a buffering proxy in front of a default configuration Gunicorn. If you exposed synchronous workers to the internet, a DOS attack would be trivial by creating a load that trickles data to the servers. For the curious, Hey is an example of this type of load.

> Some examples of behavior requiring asynchronous workers:

>Applications making long blocking calls (Ie, external web services)
Serving requests directly to the internet
Streaming requests and responses
Long polling
Web sockets
Comet

So a pretty low-effort way to test this is by switching gunicorn to asynchronous workers. This is a test of that implementation. Lets see if anything breaks/if we see fewer problems with long queue times on a build. 

# Types of changes

# How can this be tested?

 

# Checklist:
- [ ] Request reviewers


gunicorn.conf.py
```diff
@@ -1,14 +1,24 @@
 import os
 DEPLOYMENT_MODE = os.environ.get('DEPLOYMENT_MODE')
 
+worker_class = 'gevent'
```
easherma: These are gunicorn specific settings

gunicorn.conf.py
```diff
@@ -1,14 +1,24 @@
...
+workers = 4
```
easherma: There is a relationship between the number of workers, the connection pool, and how many connections end up getting opened. It's a hard relationship to express explicitly, and it takes a long time to test changes, and we don't have great ways of measuring whether certain settings are 'better' or not, but feel free to mess around.

gunicorn.conf.py
```diff
@@ -1,14 +1,24 @@
...
+# to stay under heroku limit of 20 connections
+worker_connections = 4
 
 pythonpath = "/app/joplin"
 reload = True
 #  see https://docs.gunicorn.org/en/stable/settings.html#max-requests
-max_requests = 100
+max_requests = 500
```
easherma: I upped this as it might help prevent timeouts, and the individual works _seem_ to be able to handle more simultaneous requests under this model than previously, and the startup cost of spinning up a new workers just cause it's handled 100 requests may be slowing us down (fetching route data, for example, seems to corelate to fetching around 300 requests as a single pass?)

joplin/settings.py
```diff
@@ -146,9 +146,16 @@
 #
 default_db_url = f'sqlite:///{os.path.join(PROJECT_DIR, "db.sqlite3")}'
 DATABASES = {
-    'default': dj_database_url.config(default=default_db_url),
+    'default': dj_database_url.config(default=default_db_url, engine='django_postgrespool2', conn_max_age=0),
```
easherma: this is where the connection pooling gets set up

joplin/settings.py
```diff
@@ -146,9 +146,16 @@
...
 }
 
+DATABASE_POOL_CLASS = 'sqlalchemy.pool.QueuePool'
+# https://github.com/lcd1232/django-postgrespool2#configuration
+DATABASE_POOL_ARGS = {
```
easherma: there is a relationship between this and the gunicorn settings, but I won't claim to fully understand it. The gunicorn workers are spun up and managed through a combo of the OS thread managment and the application. 

The gist here is: heroku hobby tiers allow for 20 connections, so we're assuming heroku itself may consume some of those, so we don't ever want to go more than 10 connections, but we expect around 5. 

** NOTE ** I'm _not_ totally clear on whether this applies to _each_ gunicorn worker or all of them together. 
** ALSO NOTE ** Each heroku dyno seems to spin up its own instance, but I'm _not_ sure how that affects the number of potential connections on the database. This might have been where I was running into problems running out of connections even with conservative pooling settings. 

These pooling settings refer specifically to the application, so settings here will affect what gunicorn ends up having to handle (I think). 

The 'recycle' setting  is kinda interesting, more info on that in the SQL alchemy docs themselves.

Niicck: > This might have been where I was running into problems running out of connections even with conservative pooling settings.

Do we run out of connections if we have multiple dynos running? Under what conditions do we have problems running out of connections even with conservative pooling settings?

easherma: We run out of connections if we change gunicorn to run asynchronously but don't implement any form of connection pooling. 

We run out of connections with pooling on a PR app if we set these pool size settings too high. If we set them lower, we don't run into any problems. 

We could set up some tests if we want to see where specifically we run into problems with connection pooling, but its not going to be a problem in staging or production with these settings. 

We could also implement connection pooling using pgbouncer, which I'd expect to have more predictable results.

easherma:
https://austininnovation.slack.com/archives/G82KEHEQZ/p1582661913001500?thread_ts=1582574154.012500&cid=G82KEHEQZ

Follow up: 
Merge to staging. See how it handles builds/requests. 
Up the number of allowed connections for staging & prod, because the DBs there can take more connections

PR #532, opened 2020-01-16
3388 - Make Joplin pages know their Janis URLs

briaguya:
# Description

What this does:
* Adds a Janis URL table
* Updates the Janis URL table for each page on publish

That's about it.

https://github.com/cityofaustin/techstack/issues/3388

# Types of changes

# How can this be tested?
https://joplin-pr-3388-urls.herokuapp.com/admin/
I related the Janis URLs to each page, so on save you can see "URLs" in the author interface update

# Checklist:
- [x] Request reviewers


joplin/base/models/janis_url.py
```diff
@@ -0,0 +1,199 @@
...
+    page = ParentalKey(TopicCollectionPage, related_name='janis_urls')
+    janis_url = models.ForeignKey('base.JanisUrl', verbose_name='URL', related_name='+', on_delete=models.CASCADE)
+
+
+class TopicPageJanisUrl(ClusterableModel):
+    page = ParentalKey(TopicPage, related_name='janis_urls')
+    janis_url = models.ForeignKey('base.JanisUrl', verbose_name='URL', related_name='+', on_delete=models.CASCADE)
+
+
+# @register_snippet
```
chiaberry: will it later be a snippet?

easherma: I don't think so.

joplin/base/models/official_documents_page.py
```diff
@@ -39,6 +39,7 @@ class OfficialDocumentPage(JanisBasePage):
         InlinePanel('topics', label='Topics'),
         InlinePanel('related_departments', label='Related Departments'),
         InlinePanel('official_documents', label="Documents", heading="Entries will be listed by document date (newest first)."),
+        InlinePanel('janis_urls', label='Urls'),
```
easherma: Whats the thinking behind them being in the author interface?

joplin/base/wagtail_hooks.py
```diff
@@ -300,6 +301,158 @@ def register_link_handler(features):
     features.register_link_type(InternalLinkHandler)
 
 
+# In here we're going to remake all the urls for this page
+@hooks.register('after_edit_page')
+def after_edit_page(request, page):
```
easherma: I realize this is a first attempt, but ideally what I'd like to see is the functionally broken into it's own functions (with tests), and then those functions get called on the hook here.

joplin/base/wagtail_hooks.py
```diff
@@ -300,6 +301,158 @@ def register_link_handler(features):
...
+                    official_documents_page=page if page_type == 'OfficialDocumentPage' else None,
+                    form_container=page if page_type == 'FormContainer' else None,
+                    page_type=page_type,
+                    language=language)
+                new_url.save()
+                new_urls.append(new_url)
+
+
+
+    # todo: This can definitely be less copypasta'd and more pythonic
```
easherma: I agree ;-) <3

joplin/base/models/janis_url.py
```diff
@@ -0,0 +1,199 @@
...
+    official_documents_page = models.ForeignKey("base.OfficialDocumentPage", on_delete=models.PROTECT, null=True, blank=True)
+    form_container = models.ForeignKey("base.FormContainer", on_delete=models.PROTECT, null=True, blank=True)
+
+    def __str__(self):
+        return self.url
+
+    @staticmethod
+    def get_translated_slug(language, page):
+        # todo: flesh this out
+        if language == 'es':
```
easherma: Is there a reason why we can't use this pattern? 
https://docs.wagtail.io/en/v2.7.1/advanced_topics/i18n/index.html#serving-different-languages-from-different-urls

joplin/base/models/janis_url.py
briaguya: I was thinking about making this a snippet so we could see it without going into a console, same reason I added urls to the author interface, mostly a "hey look at this PR app we can see what urls are being generated" thing

joplin/base/models/janis_url.py
briaguya: In my understanding that pattern is about actual URLs, where as these aren't actually urls but instead just strings that represent what the URL will be after a Janis build.

tl;dr - not sure how we'd be able to do that, any suggestions are very welcome

joplin/base/models/official_documents_page.py
briaguya: I wanted to see them without going into the console (and be able to show content authors what URLs are being generated)

I'm very open to removing them completely, but I was hoping to chat with @desigonz to figure out if we might want to show them somewhere

joplin/base/wagtail_hooks.py
briaguya: I agree, and I think that fits in nicely with https://github.com/cityofaustin/techstack/issues/3630, so we can just factory up pages with relationships and then make sure we get the expected URLs

joplin/base/wagtail_hooks.py
briaguya: any suggestions?

joplin/base/models/janis_url.py
easherma: We could also just display at as information as part of a panel in that case, or set up a model admin view. My guess  is that it will be weird UX to have the urls listed there and look like content authors are responsible for them when they aren't.

joplin/base/wagtail_hooks.py
easherma: I'd want to sit down with it and spend some time with it

joplin/base/models/janis_url.py
briaguya: agreed that it'd be odd having it where it is now, as I mentioned below I'm not sure if we want to show it at all, I might just remove that part from this PR completely until we have a designed decision.

joplin/base/wagtail_hooks.py
```diff
@@ -300,6 +301,158 @@ def register_link_handler(features):
...
+                        language=language)
+            new_url.save()
+            new_urls.append(new_url)
+
+        # If we're any other page type, we need to make all our urls
+        # /theme_slug/topic_collection_slug/topic_slug/page_slug/
+        # /department_slug/page_slug/
+        # /page_slug/
+        else:
+            for page_topic in page.topics.all():
```
easherma: these two querysets (looping one inside another) are going to be slow

joplin/base/wagtail_hooks.py
```diff
@@ -300,6 +301,158 @@ def register_link_handler(features):
...
+        janis_url.janis_url.delete()
+
+    # Create the new urls
+    new_urls = []
+
+    # Page type matters here
+    page_type = page._meta.object_name
+
+    # todo: not have magic language arrays hiding around the codebase
+    languages = ['en', 'es']
```
easherma: you can get these from settings.LANGUAGES

joplin/base/wagtail_hooks.py
```diff
@@ -300,6 +301,158 @@ def register_link_handler(features):
...
+@hooks.register('after_edit_page')
+def after_edit_page(request, page):
+    # Make sure we're publishing so our Janis urls reflect the live site
+    # taking this from https://github.com/wagtail/wagtail/blob/349fca66f32ecf3df6316b0cffb4e5fd0d24d6c7/wagtail/admin/views/pages.py#L228
+    publishing = bool(request.POST.get('action-publish'))
+    if not publishing:
+        return
+
+    # Clear out all old urls
+    for janis_url in page.janis_urls.all():
```
easherma: this seems wasteful. if the IA isn't changing, we shouldn't need to re-create the urls. 

I'm not 100% sure if it would be available in this hook, but as part of the form validation we get a list of changed fields. If the fields aren't related to IA, we could/should skip all this, no?

briaguya: Good ideas. This could balloon, so I made an issue for it https://github.com/cityofaustin/techstack/issues/3786

joplin/base/wagtail_hooks.py
briaguya: thought about that, this is one of the places that the disconnect between Janis and Joplin comes into play, we only have english and spanish in Janis, so it doesn't make sense to have /vi/ and /ar/ urls in here, even though we do have the ability to enter that content in Joplin.

joplin/base/wagtail_hooks.py
briaguya: How slow are you thinking? Pages generally have one topic, each topic generally has 1 topic collection, so I wasn't really worried about it.

Any suggestions for an alternative method?

briaguya:
@easherma re:

> In any case, this attempt would be about the fourth attempt we've made at spitting out janis_urls for alpha, by my count.
We've had a mix of requirements for janis urls.
So basically we went from:
* "Janis has pages with URLs and polyhierarchy support!"
* "Yeah view live links just aren't working right now"
* "View live links work on prod, but staging and pr live links point to prod too"
* "Looks like we have view live links working!"
* "We need to support departments, and pages being under them, urls need to change"
* "Janis now supports pages under department pages but view live in Joplin doesn't"
  * This is because at this point we've started trying to recreate the Janis URL creation in Joplin, so they're out of sync.
* "Joplin is tired of generating urls all over the place in different ways, let's move that somewhere and make it use decent logic"

That brings us up to the pre sitestructure days

Then, sitestructure was created as a way to make Janis ready to move to incremental builds when we get there. This is currently consumed in Janis to get a list of urls and parent/grandparent information. Since some information for a page changes based on parent/grandparent, we need to have different information when building that page at any given URL. This allows us to say, "rebuild ServicePage at department" and show the correct contextual nav information, as well as, "build it under the topic" and show the correct contextual nav information. This was previously handled using custom logic in Janis.

This has worked quite well so far, what it hasn't done a good job of handling is links. Those are a mess! We have internal links to pages and we just grab a whatever url in some places but in other places we're specifically building the url in components in Janis. I've done a little digging into it here: https://github.com/cityofaustin/techstack/issues/3787

So why the table?
* The table can replace siteStructure, and we can filter by id that way.
* We can also query and filter from the table when getting specific info for links on pages, and ideally create an API somewhat like described in https://github.com/cityofaustin/techstack/issues/3788

briaguya:
@easherma re:
> Ideally any of these ideas should pass tests before we merge the change

In order to test this we'd need to have factories for each model type, then use those factories to create instances, then make sure we update and get the same url.

That is way out of scope for what's happening here.

If you would rather do this by adding another function to JanisBasePage that returns an array of urls with relevant page information (basically an object similar to JanisUrl) that's an option, but it's not going to be any more testable than this solution until we get our models working with factories.

easherma:
> @easherma re:
> 
> > Ideally any of these ideas should pass tests before we merge the change
> 
> In order to test this we'd need to have factories for each model type, then use those factories to create instances, then make sure we update and get the same url.
> 
> That is way out of scope for what's happening here.
> 
> If you would rather do this by adding another function to JanisBasePage that returns an array of urls with relevant page information (basically an object similar to JanisUrl) that's an option, but it's not going to be any more testable than this solution until we get our models working with factories.

Although having factories to generate test data for all fields of all content types is great, we don't actually need to implement that in full before we can test this functionality. For one, we could write factories just for the IA fields. But also, we can set up test cases pretty easily with Django:
https://docs.djangoproject.com/en/3.0/topics/testing/overview/

So example test could be something like: 
* create a service page and a topic page related to service page. Run the 'get urls' function..does the expected output have 'topic page slug'/'service page slug'? What if it's related to more than one topic? etc

PR #537, opened 2020-01-23
3814 - Event content model

briaguya:
https://github.com/cityofaustin/techstack/issues/3816
https://github.com/cityofaustin/techstack/issues/3814

# Description

* Adds event pages to Joplin

# Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How can this be tested?

 
https://joplin-pr-3814-events-non-joyo.herokuapp.com/

# Checklist:
- [x] Request reviewers


joplin/api/schema.py
```diff
@@ -281,7 +393,7 @@ class Language(graphene.Enum):
     BURMESE = 'my'
 
 
-class ServicePageStepLocationBlock(graphene.ObjectType):
+class StreamFieldLocationPage(graphene.ObjectType):
```
easherma: this is a name change unrelated the event content type, no? I feel like the old name made sense. Does this need to be a part of this PR?

joplin/api/schema.py
```diff
@@ -193,6 +194,117 @@ class Meta:
         model = LocationPageRelatedServices
         interfaces = [graphene.Node]
 
+class EventPageRemoteLocation(graphene.ObjectType):
```
easherma: I remember we spoke about this verbally, but I think a docstring or extended comment here would be helpful, to explain why we aren't using the DjangoObjectType but wrote a 'custom schema'.

Also see my note below about the name of the data model

joplin/api/schema.py
```diff
@@ -193,6 +194,117 @@ class Meta:
...
+        return self.value['country']
+
+    def resolve_zip(self, info):
+        return self.value['zip']
+
+    name = graphene.String()
+
+    def resolve_name(self, info):
+        # We're doing our own translations in our model here
+        # so let's make sure the API still works as expected
```
easherma: would be great to clarify for future readers what 'as expected' means

joplin/events/models.py
```diff
@@ -0,0 +1,157 @@
...
+            ('city_location', StructBlock(
+                [
+                    ('location_page', PageChooserBlock(label="Location", page_type=[LocationPage], classname='do-not-hide')),
+                    ('additional_details_en', TextBlock(label='Any other necessary location details, such as room number [en]', required=False)),
+                    ('additional_details_es', TextBlock(label='Any other necessary location details, such as room number [es]', required=False)),
+                    ('additional_details_ar', TextBlock(label='Any other necessary location details, such as room number [ar]', required=False)),
+                    ('additional_details_vi', TextBlock(label='Any other necessary location details, such as room number [vi]', required=False)),
+                ]
+            )),
+            ('remote_location', StructBlock(
```
easherma: I realize this is a minor quibble, but I think the name 'remote location' could be kinda confusing. It makes me think that the event could be a webinar or somehow not in person. 

I know content side its described as 'non-city location', so I'd expect it to be called something like 'non_city_location', 'NonCityLocation', etc. It'l make it much easier to find in the codebase in the future.

joplin/js/editor.js
```diff
@@ -223,16 +224,20 @@ $(function() {
...
-              fieldlabel.parentNode.classList.remove('hidden')
-              if (attrFor[attrFor.length-1] !== currentLang) {
-                fieldlabel.parentNode.classList.add('hidden')
-                translatedElement.classList.remove('hidden') // only re-reveal the parent class if we find this case.
+          if (translatedElement.classList.contains('struct-block')) {
+            const fieldlabels = translatedElement.querySelectorAll('[for]');
+            fieldlabels.forEach(fieldlabel => {
+              const attrFor = fieldlabel.getAttribute('for').split('_');
+              fieldlabel.parentNode.classList.remove('hidden');
+              // Adding a failsafe to make sure we don't remove non translated fields
```
easherma: * party emoji *

joplin/api/schema.py.orig
```diff
@@ -0,0 +1,954 @@
+import django.utils.translation
+import graphene
```
chiaberry: do we need to keep the old schema? there are merge conflicts to boot

easherma: Wait, yeah I totally missed this. @briaguya whats the deal with this file?

joplin/api/schema.py
briaguya: I was going to try to reuse it for city_location on event page but forgot it was actually building an array, forgot to undo the rename when I decided not to go that route. I'm fixing it rn.

joplin/events/models.py
briaguya: @desigonz and I discussed and the plan is to have content come up with a final name, but to use remote location until we have a decision from them. I'll add some comments to the code.

joplin/api/schema.py
briaguya: addressed in new comments

joplin/api/schema.py
briaguya: addressed in new comments

joplin/api/schema.py.orig
briaguya: whoops didn't mean to check that in, good catch (too much merging)

PR #551, opened 2020-02-05
3740 with validation

easherma:
See comments on https://github.com/cityofaustin/joplin/pull/536/


joplin/templates/wagtailadmin/shared/field.html
```diff
@@ -17,5 +17,15 @@
         {% if show_help_text|default_if_none:True and field.help_text %}
             <p class="help">{{ field.help_text }}</p>
         {% endif %}
+        {% comment %} uncomment this stuff to add errors back to fields,
```
easherma: @Niicck this is why you probably weren't seeing validation errors pop up in Joplin

PR #555, opened 2020-02-06
3567 api rep error

easherma:
chased a rabbit hole into a halfway decent solution. I'll write a description tomorrow. 

This gives API output that _mostly_ matched what janis currently expects, so some tweaking would have to happen on the janis side.

But as is it restores any kind of link working in streamfields, while still handling location data in a way that matches the previous format

This started off as a fix for 3567 but is really more of a fix for the connected issue


joplin/base/migrations/0006_auto_20200205_2143.py
```diff
@@ -0,0 +1,40 @@
+# Generated by Django 2.2.9 on 2020-02-05 21:43
```
briaguya: It'd be nice to get this down to 1 migration, it's not clear what's different in the 3

joplin/api/schema.py
```diff
@@ -421,41 +449,9 @@ class Language(graphene.Enum):
     BURMESE = 'my'
 
 
-class ServicePageStepLocationBlock(graphene.ObjectType):
```
briaguya: again, wish we could use this pattern to create queryable things instead of just hardcoding our location page data in. Maybe we should make an issue for queryable streamfields and consider grapple as a possible way to make that happen

joplin/api/schema.py
```diff
@@ -192,9 +214,12 @@ class Meta:
 class LocationPageRelatedServices(DjangoObjectType):
     class Meta:
         model = LocationPageRelatedServices
+        fields = '__all__'
         interfaces = [graphene.Node]
 
 # Not final name, wating on Content team for decision
```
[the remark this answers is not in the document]
easherma: I think this is a weird thing with PEP8, its how my auto-linter is handling it I'll probably move it to a comment inside the class anyways

joplin/api/schema.py
easherma: I agree!

joplin/templates/wagtailadmin/shared/field.html
```diff
@@ -18,14 +18,14 @@
             <p class="help">{{ field.help_text }}</p>
         {% endif %}
         {% comment %} uncomment this stuff to add errors back to fields,
-        but it will show on search page
+        but it will show on search page{% endcomment %}
```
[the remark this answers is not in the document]
easherma: Ah good catch, this is unrelated to this particular issue

joplin/api/schema.py
```diff
@@ -67,7 +68,14 @@ def expand_dict_values(item):
...
     dict comprehension that expands db html on each item in a dict
     """
-    return {key: try_expand_db_html(value) for (key, value) in item.items()}
+    if isinstance(item, int):
+        return item
+    try:
+        dict_values = {key: try_expand_db_html(value) for (key, value) in item.items() if item is not isinstance(item, int)}
+        return dict_values
+    except Exception as e:
+        import pdb
```
briaguya: do we want this outside of local dev?

joplin/api/schema.py
briaguya: could you make an issue and link it?

joplin/templates/wagtailadmin/shared/field.html
```diff
@@ -27,5 +27,6 @@
     </p>
 {% endif %}
 {% endcomment %}
+
```
briaguya: was this intentional?

joplin/api/schema.py
easherma: https://github.com/cityofaustin/techstack/issues/3953

joplin/api/schema.py
easherma: good point, it shouldn't hit that point anymore but I just replaced it with a log + traceback

joplin/templates/wagtailadmin/shared/field.html
easherma: There is an endcomment right after this line, for some reason it isn't showing up in the diff?

PR #562, opened 2020-02-13
3912 wagtail 28

easherma:
Upgrades Wagtail, does some minor tweaks to maintain compatibility with embeds, and also hides the 'Reports' section of the menu, which will show you any pages you've locked from editing. 

Deployed at: https://dashboard.heroku.com/apps/joplin-pr-3912-wagtail-28

### DIFF
--- a/joplin/base/wagtail_hooks.py
+++ b/joplin/base/wagtail_hooks.py
@@ -61,7 +61,8 @@ def configure_main_menu(request, menu_items):
                      [
                          'explorer',
                          'settings',
-                         'snippets'
+                         'snippets',
+                         'reports'
                      ]
                      ]
 
--- a/joplin/settings.py
+++ b/joplin/settings.py
@@ -215,6 +215,7 @@ ALLOWED_HOSTS = [
 DEBUG_TOOLBAR = bool(strtobool(os.environ.get('DEBUG_TOOLBAR', str(False))))
 MONITOR_PERFORMANCE = bool(strtobool(os.environ.get('MONITOR_PERFORMANCE', str(False))))
 
+WAGTAILEMBEDS_RESPONSIVE_HTML = True
 
 if MONITOR_PERFORMANCE:
     MIDDLEWARE = ['silk.middleware.SilkyMiddleware'] + MIDDLEWARE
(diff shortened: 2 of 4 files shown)


robbobfrh84:
Hey @easherma, I'm curious, what is `reports` in the cinfig is related to? Also, is `WAGTAILEMBEDS_RESPONSIVE_HTML = True` new to this release or are we adding it just now?

joplin/base/wagtail_hooks.py
```diff
@@ -61,7 +61,8 @@ def configure_main_menu(request, menu_items):
                      [
                          'explorer',
                          'settings',
-                         'snippets'
+                         'snippets',
+                         'reports'
```
easherma: This is a new thing that wagtail has, which we don't do anything with, so I'm assuming we'll want to keep hiding it.

joplin/settings.py
```diff
@@ -215,6 +215,7 @@
 DEBUG_TOOLBAR = bool(strtobool(os.environ.get('DEBUG_TOOLBAR', str(False))))
 MONITOR_PERFORMANCE = bool(strtobool(os.environ.get('MONITOR_PERFORMANCE', str(False))))
 
+WAGTAILEMBEDS_RESPONSIVE_HTML = True
```
easherma: maintains previous compatibility, we still use these classes in janis (or at least they _are_ in janis)

easherma:
> Hey @easherma, I'm curious, what is `reports` in the cinfig is related to? Also, is `WAGTAILEMBEDS_RESPONSIVE_HTML = True` new to this release or are we adding it just now?

@robbobfrh84 see:
https://docs.wagtail.io/en/latest/releases/2.8.html#responsive-html-for-embeds-no-longer-added-by-default

joplin/base/wagtail_hooks.py
robbobfrh84: Curious what the reports.. report. could be really interesting.

joplin/settings.py
robbobfrh84: Gotchya!👍
... also, you responded to both of these questions in like 10 seconds.

joplin/base/wagtail_hooks.py
easherma: Check it out in https://odd-tail-bakery.herokuapp.com/admin/

admin : admin
PW: changeme

PR #563, opened 2020-02-18
3817 - schema adjustments for event list

chiaberry:
# Description

This is the joplin part of the Events List view on Janis. I needed to adjust the query so events would be a) return after a certain date, and b) in date order. I set it up so we can also return events before a certain date and the reverse date order, for when we eventually show past events. 

```
query getEventPage($id: ID, $date_Lte: Date, $date_Gte: Date) {
 	allEventPages(orderBy:"date", id:$id, date_Lte:$date_Lte, date_Gte:$date_Gte) {
    edges {
      node { }
    }
  }
}
```

adjust the `getEventPage` query by adding `orderBy: "date"` for ascending order or `orderBy: "-date"` for descending. 

date_Lte returns dates "less than or equal to" and Gte is greater than or equal to. Date are sent in "2020-03-02" format.

https://django-filter.readthedocs.io/en/master/ref/filterset.html#declaring-filterable-fields
https://docs.djangoproject.com/en/3.0/ref/models/querysets/#lte
https://github.com/graphql-python/graphene-django/blob/master/examples/cookbook/cookbook/ingredients/schema.py

# Janis instance linked to this
https://janis-3817-event-lists.netlify.com/events/

# Types of changes

# How can this be tested?

 

# Checklist:
- [ ] Request reviewers

### DIFF
--- a/joplin/api/schema.py
+++ b/joplin/api/schema.py
@@ -10,7 +10,7 @@ from graphene.types.json import JSONString
 from graphene.types.generic import GenericScalar
 from wagtail.core.fields import StreamField, RichTextField
 from wagtail.core.models import PageRevision
-from django_filters import FilterSet, OrderingFilter, DateFromToRangeFilter
+from django_filters import FilterSet, OrderingFilter
 from wagtail.core.blocks import *
 from wagtail.documents.models import Document
 from wagtail.core.rich_text import expand_db_html
@@ -238,11 +238,12 @@ class EventFilter(FilterSet):
         )
     )
 
-    # https://django-filter.readthedocs.io/en/master/guide/usage.html#the-filter
-
     class Meta:
         model = EventPage
-        fields = ['date']
+        fields = {
+            'date': ['exact', 'lte', 'gte'],
+            'live': ['exact']
+        }
 
 
 class EventPageRemoteLocation(graphene.ObjectType):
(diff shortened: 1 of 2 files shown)


joplin/api/schema.py
```diff
@@ -231,6 +231,22 @@ class Meta:
...
+    order_by = OrderingFilter(
+        fields=(
+            ('date'),
+        )
+    )
+
+    class Meta:
+        model = EventPage
+        fields = {
+            'date': ['exact', 'lte', 'gte'],
```
easherma: I think we talked about this, but adding a note/link for more info on where 'lte' and 'gte' come from, etc. so we know where to dig for more info

joplin/api/schema.py
```diff
@@ -231,6 +231,22 @@ class Meta:
         interfaces = [graphene.Node]
 
 
+class EventFilter(FilterSet):
```
easherma: Can we see an example of where this is used? 
This is probs a stretch goal (and maybe we can make a separate issue), but we should have a test that this ordering filter works correctly if we're going to  roll our own schema ( or we can use https://github.com/linuxsoftware/ls.joyous which already has test coverage )

joplin/api/schema.py
chiaberry: I put it in the PR, do you think it should be also as a comment in the code?

joplin/api/schema.py
chiaberry: example of where it's used on Janis? getAllEvents in static.config

easherma: here's an example of a simple API test case btw: 
https://github.com/cityofaustin/joplin/pull/545/files

joplin/api/schema.py
easherma: I personally wouldn't be opposed to a link to https://docs.djangoproject.com/en/3.0/ref/models/querysets/#field-lookups
, but a least a comment like 
'these 'fields' correspond to django queryset field-lookups' so future us knows where to look (and also where to _not_ look ) in docs

PR #567, opened 2020-02-19
3986-pytest-django

easherma:
Django has lots of handy testing features, mostly based on Python's unittest module, which is mostly based on a Java-style implementation of testing. 

Pytest is designed to be more functional and pythonic, with less boilerplate for setting up test cases. It's also backwards compatible with unittest and has the handy pytest-django module for doing stuff for Django apps, so this seems like a no brainer to start using to speed up test creation and [execution](https://pytest-django.readthedocs.io/en/latest/usage.html#running-tests-in-parallel-with-pytest-xdist)

# Description

# Types of changes

# How can this be tested?

 

# Checklist:
- [ ] Request reviewers


.circleci/config.yml
```diff
@@ -34,12 +34,9 @@ jobs:
             - "/usr/local/bin"
             - "/usr/local/lib/python3.7/site-packages"
           key: deps-{{ .Branch }}-{{ checksum "Pipfile.lock" }}
-      - run:
```
easherma: https://pytest-django.readthedocs.io/en/latest/database.html#django-db-setup

"The default implementation creates the test database by applying migrations and removes databases after the test run."

So we shouldn't need this extra step here

.circleci/config.yml
```diff
@@ -34,12 +34,9 @@ jobs:
...
-          name: Run migrations for test
-          command: pipenv run python joplin/manage.py migrate --noinput
       - run:
           name: Run tests
-          command: pipenv run bash .circleci/scripts/run_tests.sh
+          command: pipenv run pytest
```
easherma: there are a plethoria of other flags we can attach to this command, as well as CLI stuff so if we split up tests later we can run them in different chunks. There is also an option for parallelizing tests

joplin/base/tests/test_incremental_builds.py
```diff
@@ -56,6 +57,7 @@ def test_information_page_in_guide_page(self):
         global_page_id = Node.to_global_id(permit_guide.get_verbose_name().lower(), permit_guide.id)
         self.assertTrue(global_page_id in global_ids)
 
+    @pytest.mark.xfail
```
easherma: this lets pytest know that the test will fail, but we expect it to. 
tests can also be skipped based on a variety of criteria:
https://docs.pytest.org/en/latest/skipping.html

pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+python_paths = ./joplin/
+DJANGO_SETTINGS_MODULE = settings
+python_files = tests.py test_*.py *_tests.py
```
easherma: you can specify various file name patterns to also look for here

PR #575, opened 2020-02-24
3813 department groups adr

easherma:
I quickly wrote up an example of what an Architectural Decision Record could look like for us, to have a way of tracking decisions we make that we can go back to later per our convo a couple weeks ago.

### DIFF
--- a/docs/architecture/decisions/use-department-group-permissions-in-place-of-department-relation-models.md
+++ b/docs/architecture/decisions/use-department-group-permissions-in-place-of-department-relation-models.md
@@ -1,0 +1,39 @@
+# Use department group permissions in place of related Department models
+
+# Title
+
+## Status
+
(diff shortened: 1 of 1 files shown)


docs/architecture/decisions/use-department-group-permissions-in-place-of-department-relation-models.md
```diff
@@ -0,0 +1,38 @@
+<!-- # Title --> 
+# Use department group permissions in place of related Department models
+
+## Status
+<!-- What is the status, such as proposed, accepted, rejected, deprecated, superseded, etc.? -->
+Accepted
```
easherma: This is kinda ambiguous, but I figure since this PR is proposed as merged into the PR that will make this change, it might as well count as 'accepted'.

PR #576, opened 2020-02-24
only get live pages in site structure

chiaberry:
# Description

Unpublished pages are currently showing up in site structure, which was breaking my events pages query on janis (only getting live pages)

This PR changes it so we filter on live pages, not getting all pages. 

note from @briaguya: looks like this should handle https://github.com/cityofaustin/techstack/issues/4027

# Types of changes

# How can this be tested?

 

https://joplin-pr-3818-event-home2.herokuapp.com/

# Checklist:
- [ ] Request reviewers

### DIFF
--- a/joplin/api/schema.py
+++ b/joplin/api/schema.py
@@ -245,7 +245,6 @@ class EventFilter(FilterSet):
         model = EventPage
         fields = {
             'date': ['exact', 'lte', 'gte'],
-            'date': ['exact', 'lte', 'gte'],
             'live': ['exact'],
             'id': ['exact'],
             'canceled': ['exact'],
@@ -730,7 +729,7 @@ def get_structure_for_content_type(content_type):
 
         # Only publish event pages at the date based url
         if content_type == 'event page':
-            if page.date and page.live:
+            if page.date:
                 site_structure.append({'url': f'/event/{page.date.year}/{page.date.month}/{page.date.day}/{page.slug}/', 'type': content_type, 'id': page_global_id})
             continue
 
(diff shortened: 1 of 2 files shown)


joplin/api/schema.py
```diff
@@ -244,6 +244,7 @@ class EventFilter(FilterSet):
     class Meta:
         model = EventPage
         fields = {
+            'date': ['exact', 'lte', 'gte'],
             'date': ['exact', 'lte', 'gte'],
```
Niicck: Should this be duplicated?

joplin/api/schema.py
```diff
@@ -723,13 +724,13 @@ def get_structure_for_content_type(content_type):
...
 
-    pages = content_type_data["model"].objects.all()
+    pages = content_type_data["model"].objects.filter(live=True)
     for page in pages:
         page_global_id = graphene.Node.to_global_id(content_type_data["node"], page.id)
 
         # Only publish event pages at the date based url
         if content_type == 'event page':
-            if page.date:
+            if page.date and page.live:
```
Niicck: Is this redundant if we're already filtering by live=True?

joplin/api/schema.py
chiaberry: nope

joplin/api/schema.py
chiaberry: it is. I fixed it, thank you for keeping me honest

desigonz:
@chiaberry when this is in production, can you check to see if https://github.com/cityofaustin/techstack/issues/4027 is fixed?

PR #585, opened 2020-02-27
Gunicorn tweaks

easherma:
# Description
Performance testing is hairy, I did a lot of research on this for our situation. 
This tweaks the settings we just merged to get us a little more performance, but also with some safe defaults we can fall back onto if something breaks. 

# Types of changes

# How can this be tested?

 

# Checklist:
- [ ] Request reviewers

### DIFF
--- a/gunicorn.conf.py
+++ b/gunicorn.conf.py
@@ -5,7 +5,8 @@ DEPLOYMENT_MODE = os.environ.get('DEPLOYMENT_MODE')
 
 worker_class = 'gevent'
 workers = 2
-# to stay under heroku limit of 20 connections
+# to stay under heroku limit of 20 connections,
+# but with these should get caught by the pool
 worker_connections = 4
 preload = True
 pythonpath = "/app/joplin"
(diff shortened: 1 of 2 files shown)


gunicorn.conf.py
```diff
@@ -1,22 +1,31 @@
...
 #  see https://docs.gunicorn.org/en/stable/settings.html#max-requests
 max_requests = 500
 max_requests_jitter = 50
 timeout = 190
-preload = True
+
+if DEPLOYMENT_MODE in ("STAGING", "PRODUCTION"):
+    timeout = 190
+    worker_connections = 10
+    workers = multiprocessing.cpu_count() * 2
```
easherma: being extra safe here, we could potentially add another worker but i don't want to force timeouts by running out of connections. we could just watch staging for and see what happens over time. again with publisher coming this ought to be less of a problem

gunicorn.conf.py
```diff
@@ -1,22 +1,31 @@
...
+
+
 if DEPLOYMENT_MODE in ("LOCAL", "REVIEW"):
     timeout = 190
     loglevel = "DEBUG"
+    reload = True
```
easherma: we shouldn't have reload on in production, and staging should mirror production closely

easherma:
Tested a few builds on the branch and they worked, so I'll merge and then test staging.

PR #586, opened 2020-02-27
4016 Joplin UI: Application Banner & Primary Navigation

robbobfrh84:
Zenhub issue: https://app.zenhub.com/workspaces/techstack-5a78b88e1ce69f3510b678ef/issues/cityofaustin/techstack/4016

Design specs: https://xd.adobe.com/view/c533a077-9314-46a6-5db4-d4a7f8601e8f-b302/

Documentation: https://docs.google.com/document/d/1SdAro51Ppdi3Hirn1XkVviJ6nu89dDXtvvFKIX4T8t8/edit#heading=h.ff2tre8z4na1

# Description

# Summary 
 This PR is the first actual step toward building out our new Joplin UI. In the PR I attempted to do three things. 
1. Create a toggle-able site for two different UIs (Default: old).
  -  Allows our current UI to persist without change, 
  -  And, the new UI can be built without tangling with the old css/html/js. 
2. Organize and create a model for how we can develop the remainder of the new UI AND a best practice going forward.
3. Complete a first pass at implementing the Application Banner and Primary Navigation.

# Types of changes

# How can this be tested?

Deployed: https://joplin-pr-4016-navs.herokuapp.com/
- click the lamp in the center of the city seal. It's small, so you may have to  click around  a bit (10x10px)
- This PR only covers TWO sections of the new design. 
  - The Primary Navigation ("Side Nav")
  - The Application Banner, including User Account dropdown and menu.
- See "Design Specs" above.
- Does it match the Design?

# Technical Requirements 
- [ ] When in "Old" UI mode. These changes should not effect anything about how the UI looks or behaves. 
- [ ] Naming Convention. Consistent across all CSS Ids & classes - file and folders - Logic variable names...
- [ ] CSS: All new UI changes to CSS should include a class or id with a `coa-` prefix. This will ensure we do not change any other "under the hood" css that may use similar classes.

# Checklist:
- [x] Request reviewers


joplin/static/images/icon_down_arrow.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="12" height="7.4" viewBox="0 0 12 7.4"><defs><style>.a{fill:#1c1d1f;}</style></defs><g transform="translate(0 0)"><path class="a" d="M10.6,7.4,6,2.8,1.4,7.4,0,6,6,0l6,6Z" transform="translate(12 7.4) rotate(180)"/></g></svg>
```
easherma: Is the main reason for including this vs using the wagtail arrow already provided for the animation purposes? 

In case a little code comment might be a nice breadcrumb for future devs

(you can see available icons here:
https://joplin-staging.herokuapp.com/admin/styleguide/)

joplin/templates/joplin_UI/index.html
```diff
@@ -0,0 +1,2 @@
+{% include 'joplin_UI/application_banner.html' %}
```
easherma: awesome, love seeing the templates split up into more chunks like this

joplin/templates/joplin_UI/primary_navigation.html
```diff
@@ -0,0 +1,31 @@
...
+{% load i18n %}
+
+{% block content %}
+
+<link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
+<link href="https://fonts.googleapis.com/css?family=Medium+IBM+Plex+Sans&display=swap" rel="stylesheet">
+
+<aside id="coa-primary_navigation" class="nav-wrapper" style="display: none;" data-nav-primary>
+  <div class="inner">
+    <a href="{% url 'wagtailadmin_home' %}" class="logo" aria-label="{% trans 'Dashboard' %}">
```
robbobfrh84: Here's an example where i needed to add a template, but only had to add an Id to the top element to do all my styles. I just needed to write my SCSS to target the specific decedents of the element.

joplin/templates/joplin_UI/primary_navigation.html
```diff
@@ -0,0 +1,31 @@
+{% load staticfiles wagtailadmin_tags %}
+{% load i18n %}
+
+{% block content %}
+
+<link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
```
easherma: Correct  me if I'm wrong, but I think this import shouldn't be needed if its already in base.html?

joplin/templates/joplin_UI/primary_navigation.html
```diff
@@ -0,0 +1,31 @@
...
+<link href="https://fonts.googleapis.com/css?family=Medium+IBM+Plex+Sans&display=swap" rel="stylesheet">
```
easherma: similar to above, any particular reason not to include this on base.html for the admin? Are there some places where we don't expect to need to access this typeface?

joplin/templates/joplin_UI/primary_navigation.html
```diff
@@ -0,0 +1,31 @@
...
+
+<aside id="coa-primary_navigation" class="nav-wrapper" style="display: none;" data-nav-primary>
```
easherma: I'm curious, is this stuff kinda copied over from the wagtail template? If all were doing is adding the ID, is there a way we could do that by selecting this element with javascript and then adding the ID? Whatever we can do to avoid having to maintain extra templates would probably be good..or at least note which template in wagtail this is copied from. :-D

joplin/templates/wagtailadmin/skeleton.html
```diff
@@ -0,0 +1,78 @@
...
+  const wrapper = document.getElementById("coa-wrapper")
+  wrapper.style.paddingTop = "112px"
+  // wrapper.style.display = "none"
+  // wrapper.innerHTML = ""
+  // wrapper.innerHTML = null
+
+  const navWrapper = wrapper.querySelector('.nav-wrapper')
+  navWrapper.style.display = "none"
+  navWrapper.innerHTML = ""
+  navWrapper.innerHTML = null
```
robbobfrh84: Alright, this is the nasty part where I had to do some silly things to hide our current UI in preference to our updated UI. I not happy with it but it is temporary. And, can be removed once the new UI falls into place. I'm sure there are better ways, but it was holdin' me back. BUT, let me know if you have any ideas. its OLD UI first, so hopefully any issues will be with the new UI.

joplin/js/admin.js
```diff
@@ -1,4 +1,17 @@
-import '../css/admin.scss';
+// import '../css/admin.scss';
+
+/* 🚨
+  ONLY USED FOR JOPLIN_UI TRANSTION PERIOD
```
easherma: these notes are super helpful!

joplin/templates/wagtailadmin/skeleton.html
```diff
@@ -0,0 +1,78 @@
+{# NOTE: This template was added("copy/pasted") from the wagtail library #}
```
easherma: This is a helpful note. 

But also;
Is the only change here adding the javascript? 

You could add that using the webpack setup we have. 

Or also, it would be pretty simple to use their hook:
https://docs.wagtail.io/en/v2.8/reference/hooks.html#insert-editor-js

That way we future-proof future updates more by not having to also check all the templates. But if this is temporary it may not matter too much.

joplin/static/images/icon_down_arrow.svg
robbobfrh84: You got it!

Added a note to `applicaiton_banner.html`. And, yes the animation was the main reason. But, also this .svg from the design is a bit bolder than the built-in one, making it stand out a bit more.

robbobfrh84: Also added that link to the Joplin_UI_Dev_Notes.md Notes!

joplin/templates/joplin_UI/primary_navigation.html
robbobfrh84: You are correct. Good catch. So, what i'm gonna do is remove them from both the base.html and the primary_navigation.html. Then, put them in the skeleton.html in the header.

joplin/templates/joplin_UI/primary_navigation.html
robbobfrh84: No, I'm gonna move it. see above comment.

joplin/templates/joplin_UI/primary_navigation.html
robbobfrh84: So, my thinking here was driven by a few reasons. 

Because we want to change the style and behavior of a specific template, it's nice to have any reference to anything we've modified within our code base. 

But, beyond that, it should prevent updates from changing our html code. This happened to a few things when we updated wagtail last time, where the html of a template  changed it structure and broke our targeted css modifications. 

Without having it in our control, our entire UI is suspect to break with any minor change to the wagtail library. So, If we change it. we must own it! 

Also, in the future when somebody sees where we modified some css, they can find it within our code base easily. 

-
It's a good question, and if you think that makes since, i'll add this explanation to the documentation.

joplin/templates/wagtailadmin/skeleton.html
robbobfrh84: We've also added our custom templates in here as well as having access to the header for adding style / other dependancies. Also, there is some fun stuff that I may want to look at, like, this where they language tag is in place to be applied, which we haven't done. 
-
But, yes, the skeleton is the primary template, so this one specifically should be kept an eye on when there are any updates. 
-
So, i added some more notes in the actual file.

PR #595, opened 2020-03-03
gentle transition

easherma:
# Description

This adds the location page to the map snippet without removing the original location snippet. 

Why? Because naively nuking it causes problems with migrations and our migration testing setup, and we aren't trying to cause any more problems. 

But it does hide the locations snippets from the admin and contact pages. 

# Types of changes

# How can this be tested?

 

# Checklist:
- [ ] Request reviewers


Niicck:
What problems does it cause with our migrations?

easherma:
> What problems does it cause with our migrations?

Initially, I ran into a foreign key constraint when just dropping the snippet and replacing it with the location page. I'm thinking this is probably because loading the fixture was including  the 'old' snippet relations as well. There may also be an issue with trying to load a datadump that contains references to models that no longer exist on the target. 

I also ran into this error running migration test in general, and so I opted for pushing up a solution that still meets acceptance criteria while giving us time to follow through with removing the old stuff from the models.

PR #609, opened 2020-03-16
4116 importer

briaguya:
# Description

Allows for importing of (topic collection, topic, and information) pages from joplin 2 into joplin 3

https://github.com/cityofaustin/techstack/issues/4116

# Types of changes

# How can this be tested?

 
I wrote lots of tests, but I'd like to demo it at some point if people are down

# Checklist:
- [x] Request reviewers


joplin/importer/queries.py
```diff
@@ -0,0 +1,160 @@
+from gql import gql
+
+queries = {
+    'topiccollection': gql('''
+    query getTopicPageRevision($id: ID) {
```
briaguya: get topic collection page rename this

joplin/importer/queries.py
```diff
@@ -0,0 +1,160 @@
...
+                      }
+                    }
+                  }
+                }
+              }
+            }
+          }
+        }
+    '''),
+    'information': gql('''
```
briaguya: why are there two?

joplin/pages/topic_collection_page/factories.py
```diff
@@ -32,6 +33,64 @@ class Meta:
...
+                JanisBasePageTopicCollectionFactory.create(page=self, topic_collection=topic_collection)
+            return
+
+        # todo figure out if this is really what we want this factory to do
         if create:
             JanisBasePageTopicCollectionFactory.create_batch(2, page=self)
+
+
+
+def create_topic_collection_page_from_page_dictionary(page_dictionary, revision_id):
```
Niicck: A suggestion to help readers follow the flow: calling this something like "create_topic_collection_page_from_importer" would let us know that these factory functions are being used by the importer process.

joplin/importer/page_importer.py
```diff
@@ -0,0 +1,86 @@
...
+            self.joplin_api_endpoint = qs['CMS_API'][0]
+
+        # get a path object to play with
+        path = Path(parse_result.path)
+
+        # figure out if we're a janis preview url
+        if 'preview' in path.parts:
+            self.parse_janis_preview_url(path)
+
+    def __init__(self, url):
```
Niicck: More documentation here would help. PageImporter takes a preview_url, and parses it in order to get the information we need in order to make a page. I think it would also read clearer if parse_url wasn't a separate method and the logic was called directly in __init__.

joplin/base/views/new_page_from_modal.py
```diff
@@ -12,17 +12,31 @@
...
 import json
 
+
+def import_page_from_url(url):
+    page_importer = PageImporter(url)
+
+    revision_id = page_importer.revision_id
+    page_dictionary = page_importer.get_page_dictionary_from_revision()
+
+    page = page_importer.create_page(page_dictionary, revision_id)
```
Niicck: page_importer shouldn't need you to plug in your page_dictionary and revision_id back in, right? Aren't they already saved inside the instance?

Public methods on importer:
page_importer.get_page_dictionary_from_revision() - for the sake of testing/verification
page_importer.create_page() - creating the actual page.

Creating a page could even be a 1 liner:
page = PageImporter(preview_url).create_page()

getting the page_dictionary could be even part of the __init__ process.

briaguya: I was debating where the page_dictionary should live, I like the idea of it living in the page_importer now that I've gone the "each page has everything it needs in one query" route

I kind of like that the init for importer is just initializing it's ability to import, but a one liner could be:
 PageImporter(preview_url).get_page().create_page()
by updating get_page_dictionary_from_revision() to save to the importer itself instead of returning it out

joplin/importer/page_importer.py
briaguya: could probably get a few vars off the class that way too

PR #614, opened 2020-03-20
4076 localize docs

chiaberry:
# Description

Content needs to be able to add spanish documents along with English documents. 

This PR adds the field and also adjusts the schema so a custom resolver will return the correct document. 

Something I learned working on this: I added the fields, added some documents. Then changed the fields (adding blank is true to the spanish field) and then could not get the resolvers to work with the exact code that Nick had working on his machine. I finally realized that I needed to run migrations in order for the schema to work. 

Thanks to Eric and Nick for helping me with this PR. 

# Types of changes

# How can this be tested?

 

https://janis-4076-local-docs.netlify.com/es/health-safety/policing-in-austin/police-feedback-and-records/official-complaint-and-discipline-documents/#1
is the janis

and https://joplin-pr-4076-local-docs.herokuapp.com/admin/pages/search/ is the linked joplin.

I replaced the first entry on janis (the last one on joplin) with a fake spanish pdf, you can see it by clicking the first document on janis. other documents that were not replaced with a spanish one still have the english document.

# Checklist:
- [ ] Request reviewers


joplin/base/models/official_documents_page.py
```diff
@@ -66,7 +66,8 @@ class OfficialDocumentPageOfficialDocument(Orderable):
     authoring_office = models.CharField(verbose_name="Authoring office of document", max_length=DEFAULT_MAX_LENGTH)
     summary = models.TextField(verbose_name="Document summary")
     name = models.CharField(verbose_name="Name of Document", max_length=DEFAULT_MAX_LENGTH)
-    document = models.ForeignKey(Document, null=True, blank=False, on_delete=models.SET_NULL, related_name='+')
+    document = models.ForeignKey(Document, null=True, on_delete=models.SET_NULL, related_name='+', verbose_name="Document [en]")
+    document_es = models.ForeignKey(Document, blank=True, null=True, on_delete=models.SET_NULL, related_name='+', verbose_name="Document [es]")
```
chiaberry: the verbose name now matches the translation pattern we have for the other translated fields.

joplin/api/schema.py
```diff
@@ -506,12 +503,35 @@ class Meta:
...
         model = OfficialDocumentPageOfficialDocument
         filter_fields = ['date']
         interfaces = [graphene.Node]
 
+    def resolve_document(self, info):
+        english_doc = OfficialDocumentNodeDocument(
+            filename=self.document.filename,
+            fileSize=self.document.file_size,
+        )
+        if django.utils.translation.get_language() == 'es':
```
Niicck: A question about this - is it possible for a self.document_es to be incomplete? And some fields within a Spanish document would have to fallback to the english version?

chiaberry: as in, a file name but no file_size? the file_size comes from wagtail/django and the document the user uploads

Niicck: Yeah, I guess you couldn't save a spanish document without a document. Makes sense. Cool, ship it! :shipit:

PR #617, opened 2020-03-23
v3 api - add nodes for pages with topics and pages with topic collections

briaguya:
makes it so this works:
```
{
  allPages {
    edges {
      node {
        janisUrls
        eventpage {
          id
        }
        locationpage {
          id
        }
        departmentpage {
          id
        }
        topiccollectionpage {
          id
        }
        janisbasepagewithtopiccollections {
          topicpage {
            id
          }
        }
        janisbasepagewithtopics {
          guidepage {
            id
          }
          servicepage {
            id
          }
          informationpage {
            id
          }
          officialdocumentpage {
            id
          }
          formcontainer {
            id
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/joplin/api/schema.py
+++ b/joplin/api/schema.py
@@ -16,8 +16,8 @@ from wagtail.documents.models import Document
 from wagtail.core.rich_text import expand_db_html
 from base.models import TranslatedImage, Theme, Contact, Location, PhoneNumber, ContactDayAndDuration
 
-from pages.topic_collection_page.models import TopicCollectionPage
-from pages.topic_page.models import TopicPage, TopicPageTopPage
+from pages.topic_collection_page.models import TopicCollectionPage, JanisBasePageWithTopicCollections
+from pages.topic_page.models import TopicPage, TopicPageTopPage, JanisBasePageWithTopics
 from pages.service_page.models import ServicePage, ServicePageContact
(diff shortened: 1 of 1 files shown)


chiaberry:
oh interesting! I was anticipating something like this

```
{
  allPages {
    edges {
      node {
        janisUrls
        pages {
         id
         pageType
      }
     }
  }
 }
}
```

but now I'm grasping that's not how it would work. Thanks for this!

briaguya:
@chiaberry yeah, we will probably need to make adjustments to this to make sure we support all of our contextual nav stuff as well. 

I figure for https://github.com/cityofaustin/techstack/issues/4185 if you feel like you wished the API exposed something in a different way then let's change the API.

Niicck:
This is good PR for getting us at parity with a sitestructure-y implementation. I agree that it doesn't have to be the final step.

We got two options:
1. Keep this API structure and do cleanData stuff on Janis
2. Clean up this API structure (like into the format Chia envisions) and do the content type cleaning/formatting on the Joplin side with resolvers or something.

I feel like option 2, cleaning up the Joplin side to make a simple API, seems like the nicest way forward. It's an open conversion. I'd like to hear what @chiaberry ends up discovering/thinks is the best road for us going forward.

But for the short term, merging this in gives us a complete wip allPages API.

PR #624, opened 2020-03-25
4181 import langs

briaguya:
# Description
Makes it so when we import a page (info/service/topic/topic collection) we get english and spanish

https://github.com/cityofaustin/techstack/issues/4181

# Types of changes

# How can this be tested?

 
import an info page or a service page and see that we get english and spanish

# Checklist:
- [x] Request reviewers


joplin/pages/topic_collection_page/factories.py
```diff
@@ -59,38 +75,40 @@ def create_topic_collection_page_from_page_dictionary(page_dictionary, revision_
...
         return page
 
-
     # since we don't have a page matching the revision id, we should look
-    # for other matches, for now let's just use slug
+    # for other matches, for now let's just use the english slug
     # todo: figure out what we want the logic around importing a page with the same slug to look like
     try:
-        page = TopicCollectionPage.objects.get(slug=page_dictionary['slug'])
+        page = TopicCollectionPage.objects.get(slug=page_dictionaries['en']['slug'])
```
Niicck: non-english slug would be unique too, right?

joplin/api/schema.py
```diff
@@ -862,4 +862,4 @@ def resolve_page_revision(self, resolve_info, id=None):
         return revision
 
 
-schema = graphene.Schema(query=Query)
+schema = graphene.Schema(query=Query, auto_camelcase=False)
```
Niicck: no more auto_camelcase, right?

joplin/pages/topic_collection_page/factories.py
briaguya: we don't require a non-english slug, so it's possible to get a page without one (which our API will probably return as the english one at the moment)

joplin/api/schema.py
briaguya: it's gone

joplin/importer/page_importer.py
```diff
@@ -20,41 +21,40 @@
...
-
-        result = client.execute(queries[self.page_type], variable_values=json.dumps({'id': self.revision_id}))
-        revision_node = result['allPageRevisions']['edges'][0]['node']
-
-        # this gets us into the 'as____Page' stuff
-        page_dictionary_from_revision = next(iter(revision_node.values()))
-
-        # set the page dictionary on ourselves but also return it
-        self.page_dictionary = page_dictionary_from_revision
+        # todo: don't just hardcode lang here
```
chiaberry: why not?

briaguya: I figure this is one of those lingering todos, where we have future hope for a time when we can decide exactly which languages we want to import

for now it stays todo

joplin/pages/topic_collection_page/factories.py
Niicck: By non-english, I meant `page_dictionary['slug']`. I was just wondering because I was writing some stuff that assumed `page_dictionary['slug']` would always exist and be unique. I think that's still true.

PR #640, opened 2020-04-04
Issue 4239: move SiteSettings to the HomePage

Niicck:
# Description

Issue: https://github.com/cityofaustin/techstack/issues/4239

What has changed?
- [ ] We moved the janis base_url logic out of SiteSettings and into the HomePage.
  - This makes a lot more sense.
  - Now every page can know its URL without querying a weird JanisBranchSettings model.
    - `self.get_parent().specific.publish_url_base()`
- [ ] JanisBasePage.janis_urls() no longer returns the full URL, only the paths without the base URL. Because that's all that static.config wants.
  - janis_urls() has become something of a misnomer now, but I figured that function will change soon enough anyway so I didn't bother changing the name.
- [ ] JanisBasePage.janis_url() has been replaced by JanisBasePage.janis_publish_url(). Because that's what we really wanted to use janis_url() for in the first place. The logic is basically unchanged, but janis_publish_url() will be a full URL that includes the base_url.
  - janis_urls() => only the paths, e.g. [`/department/slug`]
  - janis_publish_url => the whole thing, e.g. `https://janis-v3-pytest.netlify.com/department/slug`
  - janis_preview_url => unchanged, it was fine.

TODO:
- [x] Update the publish() functions to no longer use SiteSettings
- [x] Purge SiteSettings

Technically a breaking change:
- removing the base_url will break the existing static.config on Janis.
  - [ ] Maybe I also need to make a v3 Janis PR?

# Types of changes

# How can this be tested?

 

# Checklist:
- [ ] Request reviewers


joplin/pages/home_page/models.py
```diff
@@ -1,15 +1,54 @@
...
     parent_page_types = []
-    # subpage_types = ['base.ServicePage', 'base.ProcessPage', 'base.InformationPage', 'base.DepartmentPage']
-    subpage_types = [
-        'service_page.ServicePage',
-        'information_page.InformationPage',
-        'department_page.DepartmentPage',
-        'topic_page.TopicPage',
-        'location_page.LocationPage',
-        'event_page.EventPage'
+    subpage_types = []
```
Niicck, line 10: For the sake of tests, we no longer want to exclude JanisBasePage as a child. At that point, I'm not sure if we need to be explicit about what subpage_types are allowed.

joplin/pages/department_page/tests.py
```diff
@@ -5,24 +5,31 @@
 
 # If we don't have any associated department group
 @pytest.mark.django_db
-def test_department_page_no_department_group():
-    page = DepartmentPageFactory.build(slug="department_slug")
+def test_department_page_no_department_group(home_page):
+    page = DepartmentPageFactory.create(slug="department_slug", parent=home_page)
```
Niicck, line 9: All the factories for these URL tests now have to be "created" instead of "built". A merely "created" factory fixture is not saved to the database, and therefore it does not have a parent. Without a parent, we cannot have a janis_publish_url, since the base_url is now being set by the parent.

joplin/pages/base_page/tests.py
```diff
@@ -119,23 +126,23 @@ def test_base_page_with_topics_with_department_not_global_urls():
...
-#     assert url == 'http://fake.base.url/global_slug_2/'
+@pytest.mark.django_db(transaction=False)
+def test_base_page_with_topics_with_topic_with_department_coa_global_urls(home_page, expected_publish_url_base):
+    page = JanisBasePageWithTopicsFactory.create(slug="global_slug_2", coa_global=True, parent=home_page)
+
+    urls = page.janis_urls()
+    janis_publish_url = page.janis_publish_url()
+
+    assert urls == ['global_slug_2/']
+    assert janis_publish_url == f'{expected_publish_url_base}/global_slug_2/'
```
Niicck, line 148: yay it works now

joplin/pages/department_page/tests.py
```diff
@@ -5,24 +5,31 @@
...
+def test_department_page_with_department_group(home_page, expected_publish_url_base):
     department = DepartmentFactory.create(add_department_page__dummy=True)
     page = department.department_page
 
+    # Adds an already saved page to a new homepage.
+    # home_page.add_child(instance=page) will not work since "page" has already been saved to the database.
+    page.move(home_page, 'last-child')
+    # Refreshing allows the "page" object to know that its parent is now "home_page".
+    # We need page.get_parent() to work in order to get page.janis_publish_url()
+    page.refresh_from_db()
```
Niicck, line 29: This took a long time to figure out.

joplin/pages/department_page/tests.py
easherma: 💜

joplin/pages/topic_collection_page/models.py
```diff
@@ -56,11 +55,13 @@ class JanisBasePageWithTopicCollections(JanisBasePage):
...
         # Add the urls for each topic collection, these pages only
         # should publish at /theme_slug/topic_collection_slug/page_slug
-        topic_collections = self.topic_collections.all()
+        urls = []
 
-        return ['{topic_collection_url}{page_slug}/'.format(
-            topic_collection_url=base_page_topic_collection.topic_collection.janis_url(), page_slug=self.slug)
-            for base_page_topic_collection in topic_collections]
+        for base_page_topic_collection in self.topic_collections.all():
+            for topic_collection_url in base_page_topic_collection.topic_collection.janis_urls():
```
chiaberry, line 61: am I reading this correctly, can a topic_collection have more than one url? Can they be under multiple themes?

joplin/pages/topic_page/models.py
```diff
@@ -26,7 +26,7 @@ class TopicPage(JanisBasePageWithTopicCollections):
     base_form_class = TopicPageForm
 
     publish_requirements = (
-        RelationPublishRequirement('top_pages'),
+        # RelationPublishRequirement('top_pages'),
```
chiaberry, line 29: are we removing this requirement?

media/documents/Cuban_Chicken_small_batch_1.htm
```diff
@@ -0,0 +1,293 @@
...
+<p><font face="Tahoma" size="2"></FONT>
+<!-- <CTProcedureRTF>{\rtf1\ansi\ansicpg1252\deff0{\fonttbl{\f0\fnil\fcharset0 Tahoma;}}
+{\*\generator Msftedit 5.41.21.2510;}\viewkind4\uc1\pard\lang1033\f0\fs20 TO PREP: Combine olive oil, garlic, orange and lime juices, cumin, oregano, chili powder, smoked paprika, salt, and pepper in a large measuring cup or cambro and blend until smooth with stick blender. Pour HALF of marinade over chicken and toss well to coat. Marinate chicken for 1-2 hours. Reserve remaining half of marinade. \par
+\par
+While chicken is marinating, toss potatoes with .25 cup olive oil, 1 tbl salt, and 1 tbl pepper and roast in 350 degree oven until tender and beginning to brown. Cool. Bring remaining marinade to a boil and simmer for 5 minutes. Set aside to cool. \par
+\par
+Remove chicken from marinade and arrange on sheet pans (discard any marinade leftover from the chicken). Bake at 350 until chicken temps at 165. Cool, then chop into bite-sized pieces. Combine chicken, potatoes, and sauce and mix well. \fs16\par
```
chiaberry, line 246: yum, freeze me some

joplin/pages/department_page/tests.py
chiaberry: duly noted!!

joplin/pages/topic_collection_page/models.py
Niicck: They cannot be under multiple themes. And a topic_collection cannot have more than one url. For now. I changed the code to iterate over the entire list (of 1) instead of just picking the first url. Functionally, the end result is the same, and our unit tests confirm that. But it leaves our janis_urls open to more flexibility in the future, should be require it.

joplin/pages/topic_page/models.py
Niicck: Yes! I'd noticed this before. But making top_pages a requirement creates a circular dependency. It hasn't really affected production yet, because a bunch of topics and topic collections were grandfathered in before we implemented publish_requirements.

We can't publish a page without a live topic.
We can't have publish a topic with a live page.

So there isn't a way to create a new topic and a new page for a new topic. Unless I'm crazy and missing something. That was something I wanted to bring up with the team.

media/documents/Cuban_Chicken_small_batch_1.htm
Niicck: oh no my secret recipe

joplin/pages/topic_page/models.py
chiaberry: you arent crazy, for me to add pages to an empty database, id have to mark the pages as "global" so I could publish them and be live and then add them to topic.  Now I could publish the topic. Then go back to page and uncheck top page/global and make the topic be its parent. snake eating its tail indeed.

media/documents/Cuban_Chicken_small_batch_1.htm
chiaberry: the 11 herbs and spices!

joplin/pages/topic_page/models.py
briaguya: I would do the same as @chiaberry when building out test data

PR #644, opened 2020-04-07
4184 event page janis urls

robbobfrh84:
ZenHub Issue: https://app.zenhub.com/workspaces/techstack-5a78b88e1ce69f3510b678ef/issues/cityofaustin/techstack/4184

Deployed: https://joplin-pr-4184-event-urls.herokuapp.com/

API Query to check guide page url: https://joplin-pr-4184-event-urls.herokuapp.com/api/graphiql#query=%7B%0A%20%20allEventPages%20%7B%0A%20%20%20%20edges%20%7B%0A%20%20%20%20%20%20node%20%7B%0A%20%20%20%20%20%20%20%20janisUrls%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D

# Description

Events should have urls now. Only when dates are added, however. 

I've  also added a couple tests as well as.... 
....went out of scope to add some query stuff to locations. 

# How can this be tested?
- Create an event
- See the allevents graphiql link above. 
- Events should NOT have any urls yet. Should be an empty array.
- Now add a *data* and a *location* to your event. 
- Now, the query should show a janis url. 

# Types of changes

# Checklist:
- [x] Request reviewers


joplin/api/schema.py
```diff
@@ -425,6 +436,12 @@ def resolve_locations(self, info):
 
         return repr_locations
 
+    def resolve_page_type(self, info):
```
briaguya, line 439: do we need this in the API? I know that janis_urls uses it but does janis?

briaguya: this is probably a good question for @chiaberry, I know we have page type resolvers in other places now too, do we need them to build janis?

joplin/pages/event_page/factories.py
```diff
@@ -0,0 +1,72 @@
...
+    except EventPage.DoesNotExist:
+        page = None
+    if page:
+        return page
+
+    # since we don't have a page matching the revision id or the slug
+    # make the combined page dictionary
+    combined_dictionary = page_dictionaries['en']
+
+    # todo: figure out where to move this so it isn't copypasta in service page
```
briaguya, line 35: just a little note to clean this up

joplin/api/schema.py
Niicck: It doesn't look like we need it anymore, but it doesn't hurt either.

joplin/pages/event_page/factories.py
robbobfrh84: Ah! Sorry, should have left a note here... 
- I wasn't sure what we all need or want for the future. I just had to add some of it to get it working. But, not sure what is/will be needed or not.

joplin/api/schema.py
briaguya: definitely more of a "can we remove something we need to copypasta around at some point" question

briaguya: not something to change now

robbobfrh84:
@briaguya `avatar images` ... 🤷‍♂️
- ...musta snuck in when i hopped on an old branch. Good catch, tho!

PR #658, opened 2020-04-09
Generating search index playground

robbobfrh84:
With some aspects of our janis build still being build out, I wanted to see if I couldn't get a page index for search workin'. 
- I modified the basepage a bit to allow for `page types` and `short descriptions` and filtered out non-`live` pages. 

```
{
  allPages(live: true) {
    edges {
      node {
        id
        title
        janisUrls
        pageType
        shortDescription
      }
    }
  }
}
```

API LINK: http://joplin-pr-base-content-type.herokuapp.com/api/graphql/?query={%20allPages(live:%20true)%20{%20edges%20{%20node%20{%20id%20title%20janisUrls%20pageType%20shortDescription%20}%20}%20}%20}

Deployed: http://joplin-pr-base-content-type.herokuapp.com


joplin/api/schema.py
```diff
@@ -173,6 +174,9 @@ class Meta:
     def resolve_janis_urls(self, info):
         return self.specific.janis_urls()
 
+    def resolve_page_type(self, info):
+        return self.content_type
+
```
robbobfrh84, line 179: @briaguya I took a look at what you were referencing yesterday. 
- With added this here, I was able to get a reference of the "content_type"/"page_type" for each page within the allPages query (link in pr description). 
-  Also, I was able to remove the page_type from the Location and Event pages.


NOTE: But, why call it `page_type` now...? I guess it should just be called `content_type` if we go this way???

PR #716, opened 2020-04-30
4215 adding related events to pages

chiaberry:
# Description

Adding a page picker to Event pages to add related pages. 

Also changing the schema so Service, Information and Location pages can show related events. 

# Types of changes

# How can this be tested?
https://joplin-pr-4215-events-api.herokuapp.com/

 
Sample Query
```
query getServicePages {
   allServicePages{
    edges {
      node {
        id
        title
        coaGlobal
        events {
           id
          slug
          title
          date
        }
      }
    }
  } 
}
```

# Checklist:
- [ ] Request reviewers


joplin/api/schema.py
```diff
@@ -400,6 +400,7 @@ class LocationPageNode(DjangoObjectType):
     page_type = graphene.String()
     janis_urls = graphene.List(graphene.String)
     owner = graphene.Field(OwnerNode)
+    events = graphene.List(lambda: EventPageNode)
```
chiaberry, line 403: https://github.com/graphql-python/graphene/issues/110#issuecomment-366515268

joplin/api/schema.py
```diff
@@ -417,6 +418,13 @@ def resolve_page_type(self, info):
     def resolve_janis_urls(self, info):
         return self.janis_urls()
 
+    def resolve_events(self, info):
```
briaguya, line 421: it looks like this is being copied around a bit, we could probably reduce copypasta by using [graphene.Interface](https://docs.graphene-python.org/en/latest/types/interfaces/) like we're doing with [DepartmentResolver](https://github.com/cityofaustin/joplin/blob/745bc1fd95cebe90c76acba9018017673cae9876/joplin/api/schema.py#L302)

chiaberry: thank you! i updated it to use graphene.Interface

joplin/api/schema.py
```diff
@@ -307,6 +307,19 @@ def resolve_departments(cls, instance, info):
...
+    events = graphene.List(lambda: EventPageNode)
+
+    @classmethod
+    def resolve_events(cls, instance, info):
+        events = []
+        event_relationships = EventPageRelatedPage.objects.filter(page__id=instance.pk).order_by('event__date')
+        for page in event_relationships:
+            if not page.event.canceled:
+                events.append(page.event)
+        return events
```
chiaberry, line 320: events are returned in date order and only if they are not canceled.

briaguya: and the filtering is happening all the way down at the [component level in janis](https://github.com/cityofaustin/janis/pull/754/files#diff-e5c49d73ef8d69841d449238bd8b2807R9), beautiful

joplin/api/schema.py
```diff
@@ -661,7 +674,7 @@ class InformationPageNode(DjangoObjectType):
     class Meta:
         model = InformationPage
         filter_fields = ['id', 'slug', 'live', 'coa_global']
-        interfaces = [graphene.Node, DepartmentResolver]
+        interfaces = [graphene.Node, DepartmentResolver, RelatedEventPageResolver]
```
briaguya, line 677: 👍

Niicck:
Also, clicking on links to events return undefined https://janis-v3-4235-janis-rel-events.netlify.app/event/2020/5/12/undefined/. I think that's a known issue. Is that getting fixed somewhere else?

PR #729, opened 2020-05-17
4380 publish bug

chiaberry:
# Description

This PR covers two of the bugs that were recently found. 

If the topic page chooser panel is open, but *nothing* has been ever selected, the relationships have length, but the dictionary is empty. We can't check if `i[DELETE]` is false because of a key error. 

Similarly for the official documents. If one of the fields in the official document chooser isn't completed, then there are errors and there is no cleaned_data. 

# Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

# How can this be tested?

https://joplin-pr-4380-pub-bug.herokuapp.com/admin/pages/22/edit/#tab-content

Add a new document, but leave out something important/required like the title or description. Update the page and the publish preflight will catch the error and not panic. 

Create an information page. Open the topic page picker but do not select anything. Update. The page should not panic. 

Doing both of the above flows on staging would result in an Internal server error. 

 

# Checklist:
- [x] Request reviewers


joplin/pages/topic_page/models.py
```diff
@@ -107,4 +107,4 @@ class JanisBasePageTopic(ClusterableModel):
     ]
 
     def __str__(self):
-        return self.topic.text
+        return self.topic.title
```
chiaberry, line 110: i discovered while trying to `print(topic)` that topics don't have text

PR #742, opened 2020-06-08
4383 group perm

chiaberry:
# Description

Adds functions to assign the django permissions to our user fixtures. Replaces `group_permissions_settings.json`. We add these permissions when the test editor is created. 

Also makes fixtures to match the themes in joplin and loads them. 

# Types of changes

# How can this be tested?

 

# Checklist:
- [x] Request reviewers


joplin/groups/fixtures/helpers/group_permissions.py
```diff
@@ -0,0 +1,28 @@
...
+
+def add_editor_permissions(group):
+    group.permissions.add(Permission.objects.get(codename="view_user"))
+    group.permissions.add(Permission.objects.get(codename="access_admin"))
+    group.permissions.add(Permission.objects.get(codename="add_document"))
+    group.permissions.add(Permission.objects.get(codename="change_document"))
+    group.permissions.add(Permission.objects.get(codename="delete_document"))
+    group.permissions.add(Permission.objects.get(codename="add_image"))
+    group.permissions.add(Permission.objects.get(codename="change_image"))
+    group.permissions.add(Permission.objects.get(codename="delete_image"))
```
Niicck, line 28: So the Editor and Moderation are groups that already exist. We want to add these permissions to those groups for every environment we run.

Using a migration script is the right call. But I don't think the logic is quite right. I could be wrong, but I don't think we want to add those permissions to pages, I think we want to add those permissions to the editor and moderator groups only. Those aren't permissions that are relevant for pages.

So like, (in pseudocode):
```
editor_group = Group.objects.filter(name="Editor")
editor_group.permissions.add(Permission.objects.get(codename="view_extra_panels"))
editor_group.permissions.add(Permission.objects.get(codename="add_contact"))
...
```

joplin/pages/service_page/fixtures/helpers/components.py
```diff
@@ -6,11 +6,20 @@
 import pages.location_page.fixtures as location_page_fixtures
 import pages.topic_page.fixtures as topic_page_fixtures
 import pages.service_page.fixtures as service_page_fixtures
+from pages.home_page.factories import HomePageFactory
 
 
 def home():
-    return HomePage.objects.first()
```
Niicck, line 12: Was a HomePage not returning? There should've been one created in migrations.

joplin/groups/fixtures/helpers/group_permissions.py
chiaberry: I'm not adding it to pages. I can't see a difference between your psuedocode and what I have above.

joplin/pages/service_page/fixtures/helpers/components.py
chiaberry: it was not. it was returning as None in my isolated user tests, which is why I check to see if exists and if not, making a new one

joplin/users/fixtures/administrative/editor_for_test_env.py
```diff
@@ -20,6 +21,7 @@ def editor_for_test_env():
     }
 
     editor_group = Group.objects.get(name="Editors")
+    add_editor_permissions(editor_group)
```
chiaberry, line 24: here is where i add the permissions to the editor fixture. I'm seeing now that I need to find a place to run the "add_editor_permissions" to the groups when they are initially made. Aren't the groups made automatically by wagtail?

joplin/pages/service_page/fixtures/helpers/components.py
chiaberry: also, this runs with tests, do migrations run before we run pytest?

joplin/users/fixtures/administrative/editor_for_test_env.py
Niicck: Yep they are created automatically by Wagtail. But I think permission could be added to them after the fact within a migration, like how you're adding page view restriction in the other migration file.

joplin/groups/fixtures/helpers/group_permissions.py
Niicck: You're right, I misread the 0002_add_pageviewpermissions.py migration file. I thought it was dealing with these group_permissions. I responded to your comment on editor_for_test_env.py

PR #746, opened 2020-06-12
4400 Prevent editors/authors from creating topics and topic collections

chiaberry:
# Description

Authors should not be able to make topics or topic collections, so that link is hidden from the create content modal. 

I registered an inclusion tag to be able to check if the user is in the editors groups or not. 

# Types of changes

# How can this be tested?

 

https://joplin-pr-4400-topic-prevent.herokuapp.com/admin/pages/search/

login as admin. click create content modal, see the option to make a topic or topic collection

login as editor, follow the same steps and you shouldnt be able to see the topic and topic collection

login: editor@austintexas.io
pw: e

# Checklist:
- [ ] Request reviewers

### DIFF
--- a/joplin/js/CreateContentModal/index.js
+++ b/joplin/js/CreateContentModal/index.js
@@ -35,6 +35,7 @@ class CreateContentModal extends Component {
       creatingContent: false,
       content_or_topic: 'content',
       missingTitle: false,
+      can_set_topics: (document.getElementById("createNewContentButton").dataset.createTopics === 'true')
     };
   }
 
@@ -190,6 +191,7 @@ class CreateContentModal extends Component {
                           this.handleContentOrTopicSelect
                         }
                         content_or_topic={this.state.content_or_topic}
+                        canSetTopics={this.state.can_set_topics}
                       />
                     )}
                     {this.state.activeStep === stepsEnum.CHOOSE_TITLE && (
--- a/joplin/js/CreateContentModal/index.js
+++ b/joplin/js/CreateContentModal/index.js
@@ -35,7 +35,7 @@ class CreateContentModal extends Component {
       creatingContent: false,
       content_or_topic: 'content',
       missingTitle: false,
-      can_set_topics: (document.getElementById("createNewContentButton").dataset.isEditor === 'false')
+      canSetTopics: JSON.parse(document.getElementById('can-create-topics').textContent).topics,
     };
   }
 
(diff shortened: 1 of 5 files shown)


joplin/js/CreateContentModal/index.js
```diff
@@ -35,6 +35,7 @@ class CreateContentModal extends Component {
       creatingContent: false,
       content_or_topic: 'content',
       missingTitle: false,
+      can_set_topics: (document.getElementById("createNewContentButton").dataset.isEditor === 'false')
```
Niicck, line 38: Reversing the logic could make this more future proof if we need to add more roles. So instead of saying everyone except an editor can set topics, we can make it more explicit by saying only moderators and superadmins can set topics. And that logic to determine who can_set_topics can be done in the joplin_tags/py.

Using the builtin "json_script" tag lets you not have to hack your own variable injection, and it'll let you use JSON spec which allows Boolean values. It also can give us a consistent way to organize how, where, and which python values are getting sent to js. 

https://docs.djangoproject.com/en/3.0/ref/templates/builtins/#json-script
ex:
`{{ style_guide_url|json_script:"style-guide-url" }}` in template
`JSON.parse(document.getElementById('style-guide-url').textContent)` in js

joplin/js/CreateContentModal/index.js
```diff
@@ -190,6 +191,7 @@ class CreateContentModal extends Component {
                           this.handleContentOrTopicSelect
                         }
                         content_or_topic={this.state.content_or_topic}
+                        canSetTopics={this.state.can_set_topics}
```
Niicck, line 194: Let's keep CamelCase for js whenever possible.

joplin/js/CreateContentModal/index.js
chiaberry: incorporated the feedback!

joplin/js/CreateContentModal/index.js
chiaberry: fixed.

joplin/templates/wagtailadmin/pages/search_header.html
```diff
@@ -1,6 +1,9 @@
 {% load i18n %}
+{% load joplin_tags %}
 
 <div class="header secondary coa-search-header ">
+    {% can_create_topics as topics %}
+    {{topics|json_script:"can-create-topics"}}
```
Niicck, line 6: The `can_create_topics` can be plugged directly into json_script

PR #772, opened 2020-07-08
allow orderBy: "-page__date" on officialDocumentCollectionDocuments

Niicck:
When the site builds, test it with:
```
query getThing($id: ID) {
  officialDocumentCollectionDocuments(officialDocumentCollection: $id, orderBy: "-page__date") {
    edges {
      node {
        id
        officialDocumentCollection {
          id
        }
        page {
          id
          title
        }
      }
    }
    pageInfo {
      hasNextPage
      endCursor
    }
}
}
```

### DIFF
--- a/joplin/pages/official_documents_page/factories.py
+++ b/joplin/pages/official_documents_page/factories.py
@@ -1,6 +1,6 @@
 import factory
 from pages.official_documents_page.models import OfficialDocumentPage, OfficialDocumentCollectionDocument
-from pages.base_page.factories import JanisBasePageFactory
+from pages.topic_page.factories import JanisBasePageWithTopicsFactory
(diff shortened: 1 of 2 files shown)


joplin/pages/official_documents_page/factories.py
```diff
@@ -28,7 +28,7 @@ class Meta:
         model = OfficialDocumentCollectionDocument
 
 
-class OfficialDocumentPageFactory(JanisBasePageFactory):
+class OfficialDocumentPageFactory(JanisBasePageWithTopicsFactory):
```
chiaberry, line 31: OfficialDocumentPage inherits from JanisBasePage, that's why the factory is JanisBasePageFactory
`OfficialDocumentPage(JanisBasePage)` 

otherwise officialdocumentpages would need topics associated with them

PR #774, opened 2020-07-13
make publish_succeeded endpoint async

Niicck:
https://github.com/cityofaustin/techstack/issues/4626

# Description

Our /publish_succeeded is now asynchronous. So even if it takes more than 30s to update all the pages, our Publisher requests won't time out. The processing of pages happens in a separate thread.

# Types of changes

# How can this be tested?
~There's an /async_test endpoint just to test that async stuff works in general on heroku.~ (It works, I removed it)
```
curl -X POST https://joplin-pr-4626-async.herokuapp.com/async_test
heroku logs -a joplin-pr-4626-async --tail
```

I'm going to try to make a Publisher build that has 200+ pages to update to make sure it works right.

# Checklist:
- [ ] Request reviewers


joplin/base/views/publish_succeeded.py
```diff
@@ -12,22 +13,40 @@
...
+            Parameters:
+                request.body.pages (list of page_data)
+                request.body.api_keys (list of str)
+            Returns:
+                Response (int): always return 200 success code.
+    '''
     body = json.loads(request.body)
     pages = body["pages"]
+    api_keys = body["api_keys"]
+    threading.Thread(target=handle_publish_success, args=(pages,api_keys)).start()
```
Niicck, line 33: By creating a separate Thread, that code can keep executing even after this function finishes and returns a response to Publisher.

PR #782, opened 2020-07-21
Issue 4402: add notification emails

Niicck:
https://github.com/cityofaustin/techstack/issues/4402

# Description

Alright. We can now have our passwords reset in Joplin. When you click on the 'forgot password' button on login, it will now actually email you a link to reset your password. And it will work.

# Types of changes

# How can this be tested?

 

Got to https://joplin-pr-4402-email.herokuapp.com/admin/login/
Create a user.
Log out and reset that user's password.

# Checklist:
- [ ] Request reviewers


joplin/settings.py
```diff
@@ -449,9 +445,13 @@ def show_toolbar(request):
...
 # The CMS_API endpoint of the current Django App for published Janis to use
 if IS_LOCAL or IS_TEST:
-    CMS_API = f"http://localhost:{JOPLIN_APP_HOST_PORT}/api/graphql"
+    # Base URL to use when referring to full URLs within the Wagtail admin backend -
+    # e.g. in notification emails. Don't include '/admin' or a trailing slash
+    BASE_URL = f'https://localhost:{JOPLIN_APP_HOST_PORT}'
+    CMS_API = f"{BASE_URL}/api/graphql"
 else:
-    CMS_API = f"https://{os.getenv('APPNAME','')}.herokuapp.com/api/graphql"
+    BASE_URL = f"https://{os.getenv('APPNAME','')}.herokuapp.com"
```
chiaberry, line 453: the default is an empty string because otherwise it would be `undefined` and break in a less graceful way?

Niicck: Yes, if we tried to do string interpolation with `None` then everything breaks.

PR #809, opened 2020-09-08
Issue 4776: implement searchIndex for Elasticsearch

Niicck:
https://github.com/cityofaustin/techstack/issues/4776

# Description

- [ ] Cleaned up our api to help with searchIndex building. Now we have inheritance. All pageType nodes inherit from JanisBasePageNode. This is good because we no longer have to duplicate resolvers for fields that exist on every content_type.
- [ ] Added a "search_summary" field and resolver. Each content_type's model will be responsible for providing the specific field they want to use as a search summary.

Janis PR: https://github.com/cityofaustin/janis/pull/853

# Types of changes

# How can this be tested?
https://janis-v3-4776-elastic.netlify.app/en/search/#page=1
This search should work the same as it did before. 
- [ ] Make sure all the page results we'd expect to see show up.
- [ ] And that those links work.
- [ ] And that the right summary/data below the link shows up for correctly for each content_type.

# Checklist:
- [ ] Request reviewers


joplin/pages/official_documents_page/models.py
```diff
@@ -68,6 +72,11 @@ class OfficialDocumentPage(JanisBasePage):
         InlinePanel('official_document_collection', label="Official document collections this document belongs to"),
     ]
 
+    class Meta:
+        indexes = [
+            models.Index(fields=['-date',]),
+        ]
```
chiaberry, line 78: 🎊

joplin/api/schema.py
```diff
@@ -179,11 +179,62 @@ class ContextualNavData(graphene.ObjectType):
...
+        return self.email
+
+    @superuser_required
+    def resolve_is_superuser(self, info):
+        return self.is_superuser
+
+
+# Add this method to a PageNode to allow authorized users to see page owner data.
+def resolve_owner_handler(self, info):
+    owner = self.owner
```
chiaberry, line 219: this is for my own edification, do some of our pages not have owners? or is it just in case so nothing breaks?

joplin/api/schema.py
```diff
@@ -194,13 +245,10 @@ def resolve_janis_urls(self, info):
         return self.specific.janis_urls()
 
     def resolve_page_type(self, info):
-        return self.content_type
+        return self.content_type.name
```
chiaberry, line 248: 👍

chiaberry (approved):
the schema.py refactoring is so good, im even angry that we didn't think of it before

joplin/api/schema.py
Niicck: Some of our pages actually don't have owners. They're legacy pages from early on in Joplin v2 that must have been created programmatically. I can't remember the story of how they got there, but they're there.

joplin/pages/official_documents_page/models.py
Niicck: In my tests, adding this index only shaved off like half a millisecond. But it's still the right thing to do, and it'll help in the long run. I documented my experiment here: https://github.com/cityofaustin/techstack/issues/4657

PR #815, opened 2020-09-24
4897 - Remove tags from images

chiaberry:
# Description

Tags aren't being used on images, so they are being removed from joplin. 

This PR removes the tags from the Image form as well as removing the tags from the image index template.

https://github.com/cityofaustin/techstack/issues/4897

Also, I ran migrations and found two that had not been applied. This is most likely due to previous oversights on my part, so I am adding them here. 

# Types of changes

- [x] New feature (non-breaking change which ~~adds~~ removes functionality)

# How can this be tested?

https://joplin-pr-4897-no-image-tags.herokuapp.com/admin/images/30/

look ma, no tags!

 

# Checklist:
- [ ] Request reviewers


joplin/base/models/translated_image.py
```diff
@@ -7,7 +7,9 @@
 
 
 class TranslatedImage(AbstractImage):
-    admin_form_fields = Image.admin_form_fields
+    # admin_form_fields = Image.admin_form_fields
+    admin_form_fields = ('title', 'file', 'collection', 'focal_point_x', 'focal_point_y', 'focal_point_width',
+                         'focal_point_height')
```
chiaberry, line 12: Image.admin_form_fields is everything above, plus tags.

Niicck (approved):
:shipit 

Out of curiosity, does the "tags" field still display in the template, even though tags are no longer part of the model? Is it just like an inert field?

chiaberry:
They do because old images have legacy tags associated with them, and since
tags are their own model, removing anything from images doesn't delete the
existing tags

robbobfrh84 (approved):

